PR #2, opened 2016-08-23
Fix typo

mfurtak:


KrauseFx:
Google has more results for `practices`, so this change looks good to me. Let's see if we can re-trigger circle 👍

KrauseFx:
Mh, seems to be stuck, as CI wasn't enabled back when this PR was submitted. I'll just submit a new PR for this change 👍

KrauseFx:
Oh, this is just in the summary of what we want to cover, not worth changing I think 👍

PR #11, opened 2016-08-25
Fix new lines in gh-pages branch circle.yml

KrauseFx:
Should fix the new line escaping in https://github.com/fastlane/docs/blob/gh-pages/circle.yml


mfurtak:
👍

PR #16, opened 2016-08-25
Restructure getting-started to have separate Apple and Android folders

mfurtak:
Since the fastlane sub-tools and other configuration info will be different between Apple and Android projects, it felt best to have two distinct areas under **getting-started**


KrauseFx:
screenshot 2016-08-25 11 42 03

This is great, I also really love the URL that is being used as a result 👍

PR #27, opened 2016-08-26
Hide the next and the previous button on each page

KrauseFx:
These buttons just don't make a lot of sense for our use-case, as it's not really gradual. This PR hides the 2 buttons.

---

**Before**
screenshot 2016-08-25 18 17 44

----

**After**
screenshot 2016-08-25 18 18 48


mfurtak:
:ship: it

PR #32, opened 2016-08-26
Add bundle install before calling fastlane

KrauseFx:
Fixes https://circleci.com/gh/fastlane/docs/103

I’m not 100% sure why this is necessary
it’s not a big deal, since this will only run on the deployment on the master branch, so it doesn’t affect the execution time for PRs


mfurtak:
👍

PR #39, opened 2016-08-29
Make use of no-highlight

KrauseFx:
Syntax highlighting was off for those commands, so we disable it


mfurtak:
TIL - will have to use that in screenshots :+1:

KrauseFx:
In the screenshots docs you mean? I'd prefer having this the default (similar to what GitHub has), but I didn't find an easy way to set it

PR #41, opened 2016-08-29
Update emoji table

KrauseFx:
The emoji table is the essence of why there are even READMEs. People decide if they want to use the software based on the choice of emojis and how appropriate there are. That's why we have to refine our choices, to become better, until we reach the limit of what you can do with emoji tables.


mfurtak:
:100:

PR #48, opened 2016-08-29
Specify dependencies in requirements.txt.

danasilver:
Having a [requirements.txt](https://pip.readthedocs.io/en/1.1/requirements.html) makes it easier to add markdown extensions (or any Python dependency), since developers can always run the same `pip install -r requirements.txt` to install the dependency locally and the [CI install command](https://github.com/fastlane/docs/blob/ae7b6a3630dce01f7ad06d7a35e9a79371bee2b6/circle.yml#L8) doesn't have to change whenever a dependency does.

The `requirements-to-freeze.txt` and `requirements.txt` setup comes from [A Better Pip Workflow](http://www.kennethreitz.org/essays/a-better-pip-workflow). I haven't used it before, but it sounds intriguing and I've been interested in giving it a try, since there isn't a good supported format like a `Gemfile.lock` or `npm-shrinkwrap.json`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # fastlane docs
 
-This repo contains all documentation for fastlane. You can find the `.md` files inside the [docs](docs) folder. 
+This repo contains all documentation for fastlane. You can find the `.md` files inside the [docs](docs) folder.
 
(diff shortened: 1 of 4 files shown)


KrauseFx:
Looks great, thanks 👍

Issue #42, opened 2016-08-29
Design goals


KrauseFx:
![screenshot 2016-08-28 20 41 30](https://cloud.githubusercontent.com/assets/869950/18040189/cfec2622-6d5f-11e6-95fe-fae186ff8aa1.png)

Design was taken from https://docs.fabric.io. We have to adapt it to be more like fastlane (https://fastlane.tools). White, with a lot of colors (the 5 at least)

**Must haves:**
- [x] Fix my hacky way to get the currenlty active sidebar element to be white, it's in `fastlane.css` the bottom part. There is some comments on why it's like this. However the menu item is too far to the right when you click on `Home` for example (only when it's active)
- [x] Sidebar should look brighter
- [x] Background color of the page should look like the one from GitHub.com
- [x] We have to link to the fastlane GitHub page in a visible way somewhere in the docs (header, footer or sidebar?)

**Secondary**
- [x] Code blocks should look more like GitHub.com
- [x] Add an `Edit on GitHub` button to every page to the template, so people can easily contribute changes, without manually finding the docs repo and the `.md` file
- [x] The table (e.g. emoji table on the landing page has too much padding, also the top row is rendered empty). Emoji table should look more like this: https://github.com/fastlane/fastlane
- [x] The page should render nicely on mobile too. Currently some images, e.g. the badges are rendered 100% width
- [x] Support anchors, so people can easily link specific sections in a document (https://facelessuser.github.io/pymdown-extensions/extensions/headeranchor/)

**Unsure**
- [x] Implement a way to fold some of the sidebar items. e.g. we want an `Advanced` button in the sidebar, with a lot of sub-pages, but we don't want to list them all the time. docs.fabric.io already does that somehow

KrauseFx:
Re anchors, I found adding this
```yml
- toc:
    anchorlink: True
```
(taken from https://github.com/tomchristie/django-rest-framework/blob/master/mkdocs.yml) helps. 

However design needs to be updated, so that headers are not blue. This is not the same as GitHub, but it's something I guess

KrauseFx:
Done 👍

PR #55, opened 2016-08-30
Extend toc.html to remove unwanted <li>s in the build.

danasilver:
Also fix top level toc items having too much padding when active. First item on #42.

Before: (from docs.fastlane.tools)

screen shot 2016-08-29 at 10 34 23 pm

After: Only visible change is no extra padding for top level active navigation items (like "Home").

screen shot 2016-08-29 at 10 34 03 pm


KrauseFx:
Beautiful 🚀

PR #59, opened 2016-08-30
"iOS" instead of "apple"

KrauseFx:
Why this change?
- Better SEO, Apple is not something someone would search for
- We don’t fully support tvOS or other platforms, but mostly focus on iOS
- When I search on docs (manually or CMD + F), I’d enter “iOS” as iOS developer
- Even though tvOS is kind of its own OS, you’re still usually an iOS developer


mfurtak:
Sounds good 👍

PR #60, opened 2016-08-30
Remove $ for running commands

KrauseFx:
The $ makes it harder for people to copy & paste commands. In the case of the snapshot example, the user couldn't copy all 3 commands before


mfurtak:
👍

PR #75, opened 2016-09-01
Update Advanced.md

KrauseFx:
- Fix number of ` for mkdcos
- Remove .fastlane section, since I think there is still some issues with some tools for that


mfurtak:
:+1:

PR #85, opened 2016-09-02
Visited links shouldn’t be purple

KrauseFx:
Visited links shown in purple is so 2000

### DIFF
--- a/docs/css/fabric.css
+++ b/docs/css/fabric.css
@@ -2443,7 +2443,6 @@ input[type="checkbox"][disabled] {
 }
 
 a {
-  color: #008BF3;
   text-decoration: none;
   cursor: pointer
 }
--- a/docs/css/fastlane.css
+++ b/docs/css/fastlane.css
@@ -25,6 +25,15 @@ blockquote {
   margin-left: 0px !important;
 }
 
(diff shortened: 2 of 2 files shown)


mfurtak:
If purple looks too retro, or doesn't fit our color scheme, I think that's OK. However, my opinion is that making `visited` have no difference in color at all is a negative for usability of the web.

KrauseFx:
But no website does a different color for visited any more, e.g. 

- GitHub
- Ruby Docs http://ruby-doc.org/
- React Docs https://facebook.github.io/react/index.html

mfurtak:
That's cool, it's just my opinion :smile:. We can do whatever we want 👍

KrauseFx:
If you're ok with it, I'd merge this 👍

mfurtak:
🆗 🚢

KrauseFx:
👍

mfurtak:
To capture some stuff that has better expressed my thinking, in case we want to reconsider later:

- https://www.smashingmagazine.com/2010/02/the-definitive-guide-to-styling-web-links/#dont-forget-about-visited-links
- http://ux.stackexchange.com/a/38007

PR #91, opened 2016-09-02
Update header to use italic fastlane

KrauseFx:


mfurtak:
Consistent rules FTW! 👍

PR #96, opened 2016-09-05
Bundler docs should mention how to install bundler

KrauseFx:


mfurtak:
This seems OK to me. As a follow-up, we're not consistent with how we format the `sudo` part of our commands.

Sometimes we're saying `sudo`, sometimes `[sudo]`. Do we have a preferred style, or is it something we're deciding more individually?

mfurtak:
👍

KrauseFx:
I'd probably go with `[sudo]`, however then people can't copy and paste the commands. I'm ok either way 👍

PR #99, opened 2016-09-05
Fix site_favicon key in mkdocs.yml.

danasilver:
Use the correct [config key](http://mkdocs.readthedocs.io/en/0.15.3/user-guide/configuration/#site_favicon) (and gets rid of the warning when building).

```
WARNING -  Config value: 'favicon'. Warning: Unrecognised configuration name: favicon
```


KrauseFx:
This is great, thanks 👍 I saw the warning before

PR #116, opened 2016-09-22
Fix typo in Crashlytics beta deployment information

johntmcintosh:
Looks like there was a small typo in the Crashlytics deployment section, where it referenced testfairy in the code snippet to show available options.


KrauseFx:
Nice catch, thanks @johntmcintosh 👍

PR #117, opened 2016-09-22
Add a section about HockeyApp beta deployment

johntmcintosh:
Here's a proposed section addition to add information about deploying beta builds to HockeyApp.


KrauseFx:
Perfect, we eventually want to extend the docs to cover more beta services, but didn't get around to it yet 👍

PR #120, opened 2016-09-26
Fix Twitter handle text in index.md.

danasilver:
Noticed this when I viewed the docs offline on my flight to SF! ✈️


KrauseFx:
👍

PR #121, opened 2016-09-26
Fix rocketship.

TKBurner:


KrauseFx:
This doesn't fix the rocketship, but remove it 😢

PR #128, opened 2016-10-03
Include information about the Snapfile to screenshot guide

KrauseFx:
Also, use `fastlane snapshot` instead of `snapshot`


mfurtak:
👍

PR #145, opened 2016-10-11
Update FAQ with instructions to enable bash/zsh tab completion for lanes names

larryonoff:
Update FAQ with instructions to enable bash/zsh tab completion for lanes names


mfurtak:
Thanks for the contribution! 👍

PR #169, opened 2016-10-26
Add missing `do` keyword from sample code

leibovic:
I'm new to fastlane, so I'm not 100% sure on the syntax here, but this fixed the issue for me.


KrauseFx:
Nice catch, thanks for finding this and submitting a PR 👍

PR #171, opened 2016-11-02
Add info about Xcode 8 Automatic Provisioning.

mpirri:


docs/codesigning/getting-started.md
```diff
@@ -61,6 +61,7 @@ Apple improved code signing a lot with the release of Xcode 8, the following has
 - No more `Fix Issue` button, instead all code signing processes run in the background and show the log right in Xcode
 - You can now specify the provisioning profile by name, instead of the UUID (Check out [xcode-project.md](xcode-project.md) for more information)
 - Improved error messages when something goes wrong. If you run into code signing errors you should always try building and signing with Xcode to get more detailed error information. (Check out [troubleshooting.md](troubleshooting.md) for more information)
+- Improved Automatic Provisioning that can create per-machine development private keys, certificates, and provisioning profiles. (Check out [xcode-project.md](xcode-project.md) for more information)
```
mfurtak: I'd say suggest using the page's title, rather than its filename, as the text that is linked; "Setting up your Xcode Project"

KrauseFx: @mfurtak I agree, this is still from the old days of having those `.md` files on GitHub directly. We do that everywhere on the `getting-started` page though, so this would keep things consistent.

mpirri: Makes sense to me; I was being consistent with the other links in this document. You think I should change all links in this document to be page-title style?

mfurtak: Ah, totally missed the prior-art. I'll :disappear: and let y'all decide what to to 👍

mpirri: All of the docs in this directory have .md files as the titles for the links.  I think I'd rather not change all of those, so I propose we leave this the way it is in the PR and we can address this later.  Make sense, @KrauseFx ?

KrauseFx: This PR is fine, I'm ok with changing the naming scheme of all links of the code signing guide in a separate PR. Probably something we should do

KrauseFx:
This PR is ready to be merged from my side

PR #184, opened 2016-11-09
Updated the docs about what sigh returns

tutipeti:
See the `sigh` specs:

```
      val = Sigh::Manager.start
      expect(val).to eq(File.expand_path("./AppStore_com.krausefx.app.mobileprovision"))
```

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -2366,7 +2366,7 @@ sigh |
 -----|----
 Supported platforms | ios
 Author | @KrauseFx
-Returns | The UDID of the profile sigh just fetched/generated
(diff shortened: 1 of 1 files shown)


KrauseFx:
This has to be updated over here, since the docs are auto-generated https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/sigh.rb#L55

However it seems like the actual fastlane action does return the UUID and not the path 👍

tutipeti:
:( Can we just pretend this never happened?

KrauseFx:
No worries, I appreciate the PR and your help 👍  😃

PR #244, opened 2016-12-21
update spaceauth docs

hjanuschka:
as discussed here: https://github.com/fastlane/fastlane/issues/7562


hjanuschka:
ci fails because of danger 😢

PR #246, opened 2016-12-23
Update continuous-integration.md

MaikoHermans:
Added Jenkins configuration with gitlab to prevent infinite loop when doing a commit/push in fastlane


KrauseFx:
Cool, thank you  🚀

PR #248, opened 2016-12-28
Explanation to use music library in screenshots

MaikoHermans:
Added explanation how to make the simulator make use of the music library so making screenshots for apps like music players can be automated as well.


KrauseFx:
Hey, thanks for submitting the PR, this looks interesting, however I have the feeling this is a very specific use-case that shouldn't necessarily be covered docs.fastlane.tools. Have you thought about putting this up as your blog post instead? So people can find it when they google for a solution for this issue.

MaikoHermans:
@KrauseFx 
You are right this is pretty specific use-case. However I was trying to get snapshot working with my music player application and couldn't find a solution anywhere for it and I was hoping you guys would have found something on it so I looked in the documentation. With this logic I figured it would be a nice addition as long as it isn't to present. 

I'll see where I can post it as well. I'm still figuring out some things with fastlane and trying stuff out. I will be writing an entire how to for CI in combination with fastlane. I will put the PR in there somewhere as well. But this might take some time since it's just one of many side projects really

KrauseFx:
Awesome, thanks for understanding 👍

PR #255, opened 2017-01-10
Add turbolinks for instant page reloads

buger:
You have one of the best documentations I ever seen, be we can make it even better with faster page reloads.

This change add small (30kb) js library called Turbolinks, which automatically will be loaded by mkdocs, which gives almost instant page reloads.

Project page https://github.com/turbolinks/turbolinks

> Turbolinks makes navigating your web application faster. Get the performance benefits of a single-page application without the added complexity of a client-side JavaScript framework. Use HTML to render your views on the server side and link to pages as usual. When you follow a link, Turbolinks automatically fetches the page, swaps in its <body>, and merges its <head>, all without incurring the cost of a full page load.

I checked on mkdoc based site, and it worked great!
But for sure require a bit more testing since you have a lot of different pages.


buger:
I believe last failed test not caused by me: "Could not set up API to Code Review site for Danger"

KrauseFx:
Yeah, don't worry about the test failure for now, for security reasons we can't run danger for external PRs.

buger:
Not sure what you mean, `mkdocs` automatically loads all files from `js` and `css` folders. If you want, I can add explicit `extra_js` option to `mkdocs` config.

KrauseFx:
Ah, alright, I didn't know that 👍

KrauseFx:
Just tested this locally, works great 👍  Thanks for your contribution

PR #256, opened 2017-01-16
Add for_lane and for_platform documentation

mfurtak:
Documents https://github.com/fastlane/fastlane/pull/7859

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -337,3 +337,52 @@ Credential felix@krausefx.com:********* added to keychain.
 fastlane-credentials remove --username felix@krausefx.com
 password has been deleted.
 ```
+
+# Control configuration by lane and by platform
+
+In general, configuration files take only the first value given for a particular configuration item. That means that for an `Appfile` like the following:
+
+```ruby
+app_identifier "com.used.id"
+app_identifier "com.ignored.id"
+```
+
+the `app_identfier` will be `"com.used.id"` and the second value will be ignored. The `for_lane` and `for_platform` configuration blocks provide a limited exception to this rule.
+
+All configuration files (Appfile, Matchfile, Screengrabfile, etc.) can use `for_lane` and `for_platform` blocks to control (and override) configuration values for those circumstances.
+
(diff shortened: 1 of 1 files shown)


docs/advanced.md
```diff
@@ -337,3 +337,52 @@ Credential felix@krausefx.com:********* added to keychain.
 fastlane-credentials remove --username felix@krausefx.com
 password has been deleted.
 ```
+
+# Control configuration by lane and by platform
+
+In general, configuration files take only the first value given for a particular configuration item. That means that for an `Appfile` like the following:
+
+```ruby
+app_identifier "com.used.id"
```
KrauseFx: Until now, I think that's how it happens, however I don't think we ever documented or added tests for the order of those. Maybe by saying that's how it works, we should have at least one automated test for this behaviour (I don't think we have one currently)

docs/advanced.md
```diff
@@ -337,3 +337,52 @@ Credential felix@krausefx.com:********* added to keychain.
...
+the `app_identfier` will be `"com.used.id"` and the second value will be ignored. The `for_lane` and `for_platform` configuration blocks provide a limited exception to this rule.
+
+All configuration files (Appfile, Matchfile, Screengrabfile, etc.) can use `for_lane` and `for_platform` blocks to control (and override) configuration values for those circumstances.
+
+`for_lane` blocks will be called when the name of lane invoked on the command line matches the one specified by the block. So, given a `Screengrabfile` like:
+
+```ruby
+locales ['en-US', 'fr-FR', 'ja-JP']
+
+for_lane :screenshots_english_only do
```
KrauseFx: 👍

docs/advanced.md
mfurtak: We do have tests for this. 👍 

https://github.com/fastlane/fastlane/blob/master/fastlane_core/spec/fixtures/ConfigFileRepeatedValueSet

This is what I was able to restore after reworking that original PR.

KrauseFx: @mfurtak Beautiful, thanks for confirming 👍

PR #259, opened 2017-01-20
Remove turbolinks, as it breaks layouts

hjanuschka:
it breaks the layout.

we already had a couple of issues, where a CMD+R/CTRL+R seem to solve it, but its not a permanent fix, removing turbo links (comming in via #255 ) - fixes it.

path to reproduce (thx @mfurtak )

https://github.com/fastlane/fastlane/issues/7930#issuecomment-274071812


hjanuschka:
not sure how to fix the danger API thing. /cc @KrauseFx

mfurtak:
Excellent! We should probably do this as a step until we can identify a way to get the best of both worlds.

⚠️ **However**, we can't merge this right now as it would be removing a library that we've identified in our legal/licenses review. Please hold this for further approval. We can reopen and merge at that time. ⚠️

PR #260, opened 2017-01-21
Fix plugins section anchor link

vinhnx:
This is a small change, but it fixes navigation error on plugin section due to capitalized anchor link.


KrauseFx:
Nice, thanks for the contribution @vinhnx 👍

PR #266, opened 2017-01-25
Remove case-only redirects

mfurtak:
Case-only redirects are causing infinite redirect loops in production, so this is a change to try to fix that.


KrauseFx:
Looks good, I guess we don't need those redirects any more

PR #269, opened 2017-01-30
Updated actions.md

markph0204:
changeling_from_git_commits; merge commit_filtering replaced 'include_merges' in the example.


KrauseFx:
Thanks for the PR, however this file is auto-updated based on the source. Please modify the `.rb` file in the `fastlane/fastlane` repo instead 👍 
Let me know if you need any help

PR #271, opened 2017-01-30
Adding an alternate download option for Xcode command line tools

lseidman:
xcode-select --install is yielding this nasty little error message (with no helpful information):

> can't install the software because it is not currently available from the software update server

I suggest we give users another option for how to download the command line tools.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -55,6 +55,9 @@ Make sure you have the latest version of the Xcode command line tools installed:
 xcode-select --install
 ```
 
+Alternatively, you can install the command line tools manually by visiting the following page:
+https://developer.apple.com/download/more/?name=command%20line%20tools
+
 ### Choose your installation method:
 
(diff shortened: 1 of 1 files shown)


lseidman:
I am running 10.11.5 and Xcode 8.2.1 by the way.

KrauseFx:
I have never heard of this error message before, when is it being shown?

lseidman:
After running `xcode-select --install` you get this alert:

https://macops.ca/images/2013/10/xcodeselect-install@2x.png

When I clicked on `Install` I was getting an error message that looks like the following:

https://i.stack.imgur.com/kQHNp.png

janpio:
Is this still a problem @lseidman?

lseidman:
@janpio we are not using fastlane anymore, so unfortunately I don't have an answer to that.

PR #272, opened 2017-01-31
Add basic troubleshooting for mkdocs installation

AliSoftware:
Feedback from experience I just had when trying to install `mkdocs` for `fastlane/docs` on El Capitan.


hjanuschka:
LGTM

PR #274, opened 2017-01-31
Adding a example of using xcargs with gym

AliSoftware:
Introducing the new syntax made possible by fastlane/fastlane#8051


hjanuschka:
LGTM

PR #300, opened 2017-03-07
Alphabetized Sigh Parameters

RaghavMangrola:
I alphabetized all of `Sighs` parameters for easier reading.


KrauseFx:
Interesting, thanks. The `Actions.md` file is auto-generated, so we can't actually change it like this. We'd have to update the generate script to sort the parameters.

PR #303, opened 2017-03-15
Add instructions for manual modification of fastlane match repo

getaaron:
As a stopgap while https://github.com/fastlane/fastlane/pull/6972 is in progress, let's provide some high-level instructions for users to update the match repo by hand.

I verified the change renders correctly locally:

image

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -386,3 +386,70 @@ end
 ```
 
 you can expect the `app_identifier` to equal `"com.forplatform.mac.forlane.release"` when invoking `fastlane mac release`.
+
+# Manually Manage the fastlane match Repo
+
+Most users can benefit from fastlane match's automatic management of the repo that stores certificates and provisioning profiles. From time to time, it may be necessary to manually change the files in this repo.
+
+For example, fastlane requires admin access to the Apple Developer account to generate the appropriate files. If you are provided with an updated certificate or profile but do not have admin access, you can manually edit the repo.
+
+### Overview
+
+Because:
+
+1. fastlane encrypts the repo, and
+2. fastlane doesn't support manual edits to the repo
+
+it's necessary to manually decrypt, then modify, then encrypt, the repo to make any changes.
+
+> These instructions presuppose you already have fastlane match configured correctly.
+
+### 🔓 Decryption Instructions
+
+The easiest way to decrypt the repo is to use the fastlane match `GitHelper` class. You can do this from an interactive Ruby console:
+
+```bash
+$ bundle console
+irb(main):001:0>
+```
+
+Then, require `match` and set the appropriate parameters:
+
+```ruby
+irb(main):001:0> require 'match'
+irb(main):002:0> git_url = 'https://github.com/fastlane/example-certificate-repo'
+=> "https://github.com/fastlane/example-certificate-repo"
+irb(main):003:0> shallow_clone = false
+=> false
+irb(main):004:0> manual_password = 'example-password'
+=> "example-password"
+```
+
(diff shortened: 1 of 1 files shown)


KrauseFx:
Thanks for taking the time to write these up. I'm actually not a big fan of people manually adding profiles to _match_. The reason for that is that very often the profiles/certificates are broken/invalid, the user doesn't know, and then later on submits an issue for _match_ assume that it's an error with _match_. By _match_ forcing the user to use _match_ to generate the profiles, it is guaranteed that they're valid.

getaaron:
Yeah, I agree it should not be encouraged. But it seemed like enough people have had *some* reason to do something manually that it was worth writing the approach down.

Feel free to add a strongly-worded warning against this. Or if you prefer we can close the PR and just point people here if they insist on doing it by hand. Up to you, no hard feelings if you don't want to merge it.

KrauseFx:
Okay, let's add a warning that this is not recommended, and put it inside a
```
<details>
<summary>XYZ</summary>
...
</details>
```
block

Yes, the test fails because of the spelling of _match_ 👍

getaaron:
@KrauseFx Not sure if I understood how you wanted to use the details block. Does https://github.com/fastlane/docs/pull/303/commits/549deba568498ee4c40cc016a044238f3072f0fd look like what you're imagining?

I also updated the *match* styling and the tests pass locally.

docs/advanced.md
```diff
@@ -406,14 +413,14 @@ it's necessary to manually decrypt, then modify, then encrypt, the repo to make
...
-The easiest way to decrypt the repo is to use the fastlane match `GitHelper` class. You can do this from an interactive Ruby console:
+The easiest way to decrypt the repo is to use the fastlane *match* `GitHelper` class. You can do this from an interactive Ruby console:
 
 ```bash
 $ bundle console
 irb(main):001:0>
 ```
 
-Then, require `match` and set the appropriate parameters:
+Then, require *match* and set the appropriate parameters:
```
KrauseFx: Please use `_match_` 👍

getaaron:
@KrauseFx I updated to use `_match_` instead of `*match*`.

I also started working on a test to enforce this, but `include?` gives an undesired positive result on `**fastlane**` in `docs/best-practices/continuous-integration.md`. I made an [ugly regex](http://rubular.com/r/k4CY0umMG3) and then thought about bringing in a Markdown parser but decided it might not be worth the effort :)

docs/advanced.md
```diff
@@ -386,3 +386,79 @@ end
 ```
 
 you can expect the `app_identifier` to equal `"com.forplatform.mac.forlane.release"` when invoking `fastlane mac release`.
+
+# Manually Manage the fastlane match Repo
+
+Most users can benefit from _match_'s automatic management of the repo that stores certificates and provisioning profiles. From time to time, it may be necessary to manually change the files in this repo.
+
+For example, fastlane requires admin access to the Apple Developer account to generate the appropriate files. If you are provided with an updated certificate or profile but do not have admin access, you can manually edit the repo.
```
KrauseFx: Should be `_fastlane_`

docs/advanced.md
```diff
@@ -386,3 +386,79 @@ end
...
+<details>
+<summary>Instructions</summary>
+
+
+
+### Overview
+
+Because:
+
+1. fastlane encrypts the repo, and
```
KrauseFx: Same here, _fastlane_

KrauseFx:
Beautiful, thanks for updating the PR 👍

getaaron:
@KrauseFx It doesn't look like the automatic deploy worked. [Last push to `gh-pages`](https://github.com/fastlane/docs/blob/gh-pages/advanced/index.html) was Feb 4 (and I don't see it on https://docs.fastlane.tools/advanced/)

PR #305, opened 2017-03-20
Fix CI

hjanuschka:
* Update Gemfile.lock
  * update README (new tables)
  * always use latest fastlane (bundle install) on CI run.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -8,6 +8,7 @@ dependencies:
     - pip install -r requirements.txt
 test:
   override:
+    - bundle update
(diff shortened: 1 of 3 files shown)


circle.yml
```diff
@@ -8,6 +8,7 @@ dependencies:
     - pip install -r requirements.txt
 test:
   override:
+    - bundle update
```
KrauseFx: Ah that was the error then I guess

hjanuschka: exactly.

PR #309, opened 2017-03-21
add no javascript redir info

muescha:


KrauseFx:
Thanks for the PR. What is this for? Only for browsers with no Javascript?

muescha:
yes. when javascript is disabled in browser

muescha:
BTW: no need for `cla` in this repo?

mfurtak:
Good point, something is not right with that. Please comment once more?

muescha:
📖

muescha:
✅

KrauseFx:
I'm not sure, are there still browsers/people with no Javascript and do we have/want to support it?

KrauseFx:
I'm sure fastlane.tools is pretty broken without JS too

muescha:
yes a little - as i see in a first overview: 

- code snippets are not visible 
- search is not working

muescha:
but i think for search engines it would be better to do a `301 Moved Permanently` redirect ... but you can't do this with plain html / javascript :( 

some old information mention that google also count `<head><meta http-equiv="refresh" content="0; url=#{to}"></head>` like a `301`

KrauseFx:
Thanks for clarifying. For now, I think we'll just leave it as is 👍 :)

PR #312, opened 2017-03-21
fix redirect from pointing to `/##hash` 

muescha:
https://docs.fastlane.tools/codesigning/GettingStarted#using-cert-and-sigh

redirects to with double hash:

https://docs.fastlane.tools/codesigning/getting-started/##using-cert-and-sigh

(and don't jump to section with `##`)

### DIFF
--- a/scripts/ci/generate_redirects.rb
+++ b/scripts/ci/generate_redirects.rb
@@ -9,7 +9,7 @@ docs_redirects.each do |from, to|
 
   full_path = File.join(from, "index.html")
   to = "/#{to}" unless to.start_with?("/")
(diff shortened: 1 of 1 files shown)


KrauseFx:
I don't fully understand, could you explain what was wrong until now?

muescha:
now it redirects:

https://docs.fastlane.tools/codesigning/getting-started/##using-cert-and-sigh
`##` -> the browser don't jump to the section 

but it should redirect to:

https://docs.fastlane.tools/codesigning/getting-started/#using-cert-and-sigh

`#` -> the browser jump to the section

KrauseFx:
Yeah, now I understand, perfect, thanks for the fix 🚀

PR #321, opened 2017-04-05
Add another code sign gotcha to the Common Issues

shubrich:
Make sure that the private key can be accessed by command line tools on the build machine.

This trips me up every time I have to re-create the production certificate plus all provisioning certificates and move them over to the build machine.


shubrich:
I signed it!

PR #347, opened 2017-06-17
Add initial CSS for plugin scores

KrauseFx:
Via https://github.com/fastlane/fastlane/pull/9511

screenshot 2017-06-16 07 04 28

### DIFF
--- a/docs/css/fastlane.css
+++ b/docs/css/fastlane.css
@@ -51,6 +51,32 @@ h3, h4, h5, h6 {
   display: none;
 }
 
+/* For plugins quality index */
+.quality-index {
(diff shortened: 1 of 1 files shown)


docs/css/fastlane.css
```diff
@@ -51,6 +51,32 @@ h3, h4, h5, h6 {
   display: none;
 }
 
+/* For plugins quality index */
+.quality-index {
+  height: 50px;
```
KrauseFx: so many random numbers, right @taquitos?

taquitos: only the best random numbers allowed.

PR #360, opened 2017-07-12
Updates to better call out what platforms are supported

taquitos:


taquitos:
@KrauseFx I'm somewhat happy you're ok with this.

PR #407, opened 2017-08-31
Implement new action details

KrauseFx:
- Add support for running code sample tests for all sub-files (see [blog post](https://krausefx.com/blog/writing-automated-tests-for-your-documentation))
- Add support to skip tests for non Fastfile tests (different code samples, e.g. `Gymfile` code samples) by using the ``\`ruby-skip-tests notations at the beginning of a Ruby code block, this gets converted to just ```ruby during the deployment


fastlane/Fastfile
```diff
@@ -134,6 +134,10 @@ end
 # Run the code samples, yo
 private_lane :ensure_code_samples do
   UI.message "🕗  Verifying all code samples work with the latest fastlane release"
-  test_sample_code(path: File.expand_path("docs/actions.md"))
+  all_content = ""
+  Dir["docs/actions/*.md"].each do |tool_sub_page_path|
+    all_content += File.read(tool_sub_page_path)
```
ohayon: Is it okay that this is going to concatenate everything together without a space or new line or anything? I think I am just missing some context, but I definitely trust youre doing the right thing here.

KrauseFx: Nice catch and great question. In this case it shouldn't make any difference, as we only use the resulting string to find `\``ruby in the text to then run code samples. So even if the last and the first thing of each page is a code sample (which isn't the case), it will work 💯

PR #435, opened 2017-10-13
Add default values to gym parameters

revolter:
This should be added for the other tools as well!

### DIFF
--- a/docs/actions/gym.md
+++ b/docs/actions/gym.md
@@ -283,18 +283,18 @@ gym(
 
 **Parameters**
 
-Key | Description
-----|------------
+Key | Description | Default
+----|-------------|--------
   `workspace` | Path to the workspace file
   `project` | Path to the project file
   `scheme` | The project's scheme. Make sure it's marked as `Shared`
-  `clean` | Should the project be cleaned before building it?
-  `output_directory` | The directory in which the ipa file should be stored in
+  `clean` | Should the project be cleaned before building it? | false
+  `output_directory` | The directory in which the ipa file should be stored in | .
   `output_name` | The name of the resulting ipa file
   `configuration` | The configuration to use when building the app. Defaults to 'Release'
-  `silent` | Hide all information that's not necessary while building
+  `silent` | Hide all information that's not necessary while building | false
   `codesigning_identity` | The name of the code signing identity to use. It has to match the name exactly. e.g. 'iPhone Distribution: SunApps GmbH'
-  `skip_package_ipa` | Should we skip packaging the ipa?
+  `skip_package_ipa` | Should we skip packaging the ipa? | false
   `include_symbols` | Should the ipa file include symbols?
   `include_bitcode` | Should the ipa file include bitcode?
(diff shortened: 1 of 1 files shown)


KrauseFx:
Those files are auto-generated actually, we can't modify it here, check out the fastlane main repo for the generation code 👍

KrauseFx:
https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/documentation/markdown_docs_generator.rb and https://github.com/fastlane/fastlane/blob/master/fastlane/lib/assets/ActionDetails.md.erb

revolter:
I thought so, and I actively searched for something like this, but couldn't find it, neither in this repo's README nor in the main repo's CONTRIBUTING.

revolter:
How do I build the docs?

janpio:
_That_ is indeed documented at https://github.com/fastlane/docs#fastlane-docs

The other points you mentioned... not so much. Have the same problem all the time, I think it is mostly because stuff is being moved from the repo READMEs to the docs right now.

Feel free to do write stuff down you learn though, I at least would appreciate it.

revolter:
@janpio, are you referring to `mkdocs serve`? Or am I missing something?

Isn't that building the HTML pages from the Markdown files? I wanted to build the Markdown files from the Ruby source files.

I, having only one PR made, find it quite difficult to contribute to Fastlane, due to the missing or hard to find instructions on how to do stuff, and it being split into multiple documents like https://github.com/fastlane/fastlane/blob/master/CONTRIBUTING.md and https://github.com/fastlane/fastlane/blob/master/CORE_CONTRIBUTOR.md for example.

I finally managed (after hours of searching docs) to do it by commenting out these lines:

https://github.com/fastlane/fastlane/blob/c7c6a5a4f5957a979245cc1b76314cdd0309f33e/fastlane/Fastfile#L205-L214

and running `bundle exec fastlane update_docs`, and a couple of minutes to figure out (I didn't read the verbose logs at first) that the generated Markdown files were in the temp folder.

Also, one thing I don't understand is that running `fastlane` automatically generates a `README.md` file with the available lanes, but I couldn't find it in `fastlane/fastlane` repo.

janpio:
> are you referring to mkdocs serve? Or am I missing something?

That was the only I myself knew about - seems it wasn't enough.

Now that you know how it works, please write it down somewhere. As it is about how to work on the docs, it porbably belongs in a .md file in the docs repo I would say - but that is also just a guess.

janpio:
> Also, one thing I don't understand is that running fastlane automatically generates a README.md file with the available lanes, but I couldn't find it in fastlane/fastlane repo.

That should probably be an issue for the fastlane/fastlane repo then, that this file is missing. Or are you looking for where this file is generated?

revolter:
> Now that you know how it works, please write it down somewhere.

I don't think commenting out lines in a file and running it is a good solution though.

> That should probably be an issue for the fastlane/fastlane repo then, that this file is missing. Or are you looking for where this file is generated?

I think it's not generated at all, but I don't understand why.

PR #445, opened 2017-11-05
Action guide pt. 1 (was: Plugin guide pt. 1)

jdee:
This is a first installment. It includes sections on:

- FastlaneCore::Configuration
- FastlaneCore::UI
- Shelling out
- Using other actions

These are topics that come up from time to time among plugin authors. I discovered a fair bit of what's here just by reading code. Before I discovered `FastlaneCore::ConfigItem#auto_convert_value` I had code like this in a number of places:

```ruby
params[:array_param].split(",") if params[:array_param].kind_of? String
```

To say nothing of

```ruby
UI.user_error!("Bad mojo") and return unless success?
```

It's possible that the content here exists in other places, but I haven't found all of it.

I still want to add an explanation of what's in the plugin template and how to use it (CI setup, coverage, wrestling RuboCop, etc.).

There's a lot of content here. Maybe it should have its own page. For now, I put it under the Advanced section of Create Your Own Plugin.


docs/plugins/create-plugin.md
```diff
@@ -154,3 +154,322 @@ gem "fastlane-plugin-[plugin_name]", git: "https://github.com/[user]/[plugin_nam
 #### Multiple actions in one plugin
 
 Let's assume you work on a _fastlane_ plugin for project management software. You could call it `fastlane-plugin-pm` and it may contain any number of actions and helpers, just add them to your `actions` folder. Make sure to mention the available actions in your plugin's `README.md`.
+
+#### Using FastlaneCore::Configuration
+
+Note that it is not necessary to `require "fastlane_core"` in a plugin.
```
KrauseFx: Can we just remove this line?

docs/plugins/create-plugin.md
```diff
@@ -154,3 +154,322 @@ gem "fastlane-plugin-[plugin_name]", git: "https://github.com/[user]/[plugin_nam
...
+```bash
+MY_NEW_ACTION_COMMIT=true
+MY_NEW_ACTION_COMMIT=false
+MY_NEW_ACTION_COMMIT=yes
+MY_NEW_ACTION_COMMIT=no
+```
+
+These values may also be passed in all caps, e.g. `MY_NEW_ACTION_COMMIT=YES`.
+
+##### Array parameters
```
KrauseFx: This is all really awesome! However those are not plugin specific docs, right? But also when building your own actions, or when contributing to the built-in actions, right?

KrauseFx:
That's awesome, thanks so much for this. Do you think we should move this into a more general area of the docs, or should we keep it inside the plugin docs?

docs/plugins/create-plugin.md
jdee: I agree. This is more general than plugins.

jdee:
I'd like to add some plugin-specific stuff about what's in the template.

Other than that, this is pretty general. Maybe this should go into a new section under Advanced, like "Building Actions."

KrauseFx:
Sounds great, let me know when it's ready for review

jdee:
@KrauseFx and anyone else, this is all I have for actions. I'll submit a separate PR for the plugin template stuff. This is pretty big. Maybe it should have its own page. There is quite a bit on configuration. Maybe that wants its own page. I think it's ok here as well.

svanzoest:
@jdee could you cover how to approach fastlane/fastlane#10769 as well? It would be much appreciated. I love this. :) Very helpful.

jdee:
@svanzoest Thanks for the feedback. I commented on your issue, which I missed before.

There is something on lane context in the Advanced section: https://docs.fastlane.tools/advanced/#lane-context. Do you think more is needed? If this grows into a separate page, maybe some of the Advanced section could move there.

svanzoest:
@jdee I guess it wasn't clear to me how lane_context, action parameters and action environment variables interplay at all. It looks like I need to wire them up all myself?  Also, what data types are okay to pass through there? I assume passing through full ruby objects is not recommended?

Your boolean update with setting `is_string: false` is gold, as that took me a few minutes to work out, as I was trying to stick to what looks like the new API with the `type` argument to `ConfigItem`.
Also using integers for `ConfigItem.type` seems to break entering them via the fastlane prompts, but maybe your autoconvert mention addresses that.

docs/advanced.md
```diff
@@ -622,3 +622,541 @@ If you want to access those values from within your `Fastfile` use
...
+
+```Ruby
+`git commit -aqm#{Shellwords.escape commit_message}`
+```
+
+```Ruby
+system "cat #{path.shellescape}"
+```
+
+## Calling other actions
```
jdee: @ohayon I tried to summarize here the discussion from https://github.com/fastlane/fastlane/pull/10675. It's deliberately a bit vague. Any thoughts on what to say here?

janpio:
Great content. 

I think this should get merged, edited down or moved around later. For now it's important to get it in, so one can get an overview of what is where when it is merged to think about how to structure this in an improved way.

jdee:
My pleasure.

jdee:
@svanzoest I realized I never answered your question. The `lane_context` is just a Ruby Hash that lives for the lifetime of the lane. It's common to use symbols as keys in a Hash. The keys you may expect to find populated are listed in the `Actions::SharedValues` module. This is deliberately extensible to let you add things your action may need to communicate to other actions. Any Ruby reference can be passed in the lane context. It's not serialized. It's no more expensive to pass a custom class than to pass a String. But consider how much memory is used by anything you pass, since that memory will be taken up for the duration of the lane. If you need to add a value:

```Ruby
module Fastlane
  module Actions
    # define a new key in SharedValues, which will be available everywhere else
    module SharedValues
      MY_NEW_SHARED_VALUE = :MY_NEW_SHARED_VALUE
    end

    def run(params)
      # ...
      lane_context[SharedValues::MY_NEW_SHARED_VALUE] = shared_thing
```

Now all actions, including any other actions in your plugin, say, will have the definition of `SharedValues::MY_NEW_SHARED_VALUE` and can access it as `lane_context[SharedValues::MY_NEW_SHARED_VALUE]`. So yes, you have to wire it up yourself, but there's not really anything to it.

I don't think there's any real connection between the params to an action and the lane context. I don't think there's a way to say, "Map a certain parameter from this action to something in the lane context," other than writing:

```Ruby
lane_context[SharedValues::MY_NEW_SHARED_VALUE] = params[:something]
```

We can add some content like this to the docs as well as some stuff on unit testing.

I'm not sure what the problem might be with Integers. Maybe you should open an issue if you can't get that working.

svanzoest:
@jdee thank you for that write up.  

I think it is useful to know that they are expected to be distinct things, as `ConfigItem` has the environment parameter, so I wasn't clear if there maybe some logic behind the scenes that I may need to be aware of.

I got around the integer issue, by just using a string instead and casting to an integer when needed. As I work with plugins more and I do find that something is unclear or could be improved I will create a separate issue. A lot of this documentation really helps understand the intent and scope of each of these items . As someone who was new to fastlane, it is very helpful to see these things written up this way, so you are not second guessing. It will really help to minimize ramp up time for new developers.

I am very happy to see this PR merged in.

PR #448, opened 2017-11-10
Add documentation on how to detect screengrab at runtime.

dustinrwh:
Add Advanced screengrab section to screengrab.md for both launch arguments and detecting screengrab at runtime.

Worked with @joshdholtz on figuring out a way to [detect screengrab at runtime](https://github.com/fastlane/fastlane/issues/10863). Figured this would be useful to others as well 👍


dustinrwh:
I signed it!

PR #460, opened 2017-11-20
Improve docs for "Lane Context"

janpio:
Text didn't make too much sense, moved stuff around so it does and added some bits.


taquitos:
🎈:goat:

PR #461, opened 2017-11-21
[Bot] Update docs for latest fastlane release (actions.md, available-plugins.md) 🚀

taquitos:
Auto-generated by _fastlane_ bot 🤖

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -1,3 +1,8 @@
+<!--
+This file is auto-generated and will be re-generated every time the docs are updated.
+To modify it, go to its source at https://github.com/fastlane/fastlane.
+-->
+
 # fastlane actions
 
 This page contains a list of all built-in fastlane actions and their available options.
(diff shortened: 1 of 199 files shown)


docs/actions.md
```diff
@@ -1,3 +1,8 @@
+<!--
+This file is auto-generated and will be re-generated every time the docs are updated.
+To modify it, go to its source at https://github.com/fastlane/fastlane.
+-->
+
```
revolter: It looks like my PR works <3

revolter:
The comment is missing from `docs/plugins/available-plugins.md`. Do you know which part of the code generates it, so I can open a new PR?

taquitos:
@revolter it's from https://github.com/fastlane/fastlane/blob/master/fastlane/actions/plugin_scores.rb
but specifically, that file is passed in as a param: `params[:output_path]`

revolter:
@taquitos, I think it's because https://github.com/fastlane/fastlane/pull/10910, so I opened #462.

janpio:
(Triggered a retest on circleci with current result from #462 - let's see who is right in there 🗡 )

revolter:
@janpio,

1. How?
2. It would most probably work, because the case in which I personally think it doesn't (returns ok when it shouldn't) is:

```
test

<!-- comment -->
```

revolter:
Shouldn't this be rebased on top of `master`?

janpio:
Oh right... sorry. I ran the same tests again then ;)
Do that, then it will automatically test again.

taquitos:
Sent a new PR and merged that already

PR #462, opened 2017-11-21
Fix test lane crashing for HTML comments

revolter:
Introduced by https://github.com/fastlane/fastlane/pull/10910.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -62,7 +62,7 @@ private_lane :ensure_headers do
   errors = []
   Dir["docs/**/*.md"].each do |current_file|
     next if exceptions.include?(current_file)
-    errors << "File '#{current_file}' doesn't start with #" unless File.read(current_file) =~ /^#|<!--/
+    errors << "File '#{current_file}' doesn't start with #" unless File.read(current_file) =~ /^(#|<!--)/
   end
   errors.each { |e| UI.error(e) }
   UI.user_error!("At least one file doesn't start with header one (# or <!--), the first line must always be # title or <!-- comment") unless errors.empty?
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -62,7 +62,7 @@ private_lane :ensure_headers do
   errors = []
   Dir["docs/**/*.md"].each do |current_file|
     next if exceptions.include?(current_file)
-    errors << "File '#{current_file}' doesn't start with #" unless File.read(current_file).start_with?("#")
+    errors << "File '#{current_file}' doesn't start with #" unless File.read(current_file) =~ /^#|<!--/
   end
   errors.each { |e| UI.error(e) }
   UI.user_error!("At least one file doesn't start with header one (#), the first line must always be # title") unless errors.empty?
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -65,7 +65,7 @@ private_lane :ensure_headers do
     errors << "File '#{current_file}' doesn't start with #" unless File.read(current_file) =~ /^#|<!--/
   end
   errors.each { |e| UI.error(e) }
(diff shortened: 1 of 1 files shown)


taquitos:
So close, I tested with:
```
something here

<!-- Thanks for contributing to _fastlane_! Before you submit your pull request, please make sure to check the following boxes by putting an x in the [ ] -->
```

and it still passed

I think the regex is not limited to the first line

revolter:
@taquitos, Obviously, sorry, could you please test it now?

taquitos:
I think you want `\A\s*^(#|<!--)`
😄

revolter:
[`\A(#|<!--)`](http://rubular.com/r/M9bLOPyDOz) would suffice, and I don't think having superfluous whitespace at the beginning of the file is something that we want. Didn't know that Ruby regex differs for `^`. What do you say?

taquitos:
Totally fine to nix the white space, I had to look this up: 
`^` only works for the start of each string. Each line of the file is a new string, so the regex will continue to look at each line. `\A` works because it only operates on the first string.

revolter:
Rubular says that `^` matches the `Start of line` while `\A` matches the `Start of string`. In our case, the matching is done agains a single string that contains all the file's content (`File.read(current_file)`), and not on every line of it.

taquitos:
Right, got my terminology mixed up. Ok, that's great, so either will work in this case. I was using a regex tester, so multi-line must be on by default there.

taquitos:
🎈:goat:

revolter:
Wait, we didn't come to an agreement so I didn't commit the `\A` instead of `^` 😟  Could you manually fix it?

taquitos:
No no, we did :) I agree with you: `^` works here :)

revolter:
No, `^` [doesn't work](http://rubular.com/r/mfd7I3Ebu1), because it basically works like this:

> match # or <!--" that sits at the beginning of a **line**

while `\A` works like this:

> match # or <!-- that sits at the beginning of the entire specified **string**.

My [previous comment](https://github.com/fastlane/docs/pull/462#issuecomment-346087184) was just to fix [your comment](https://github.com/fastlane/docs/pull/462#issuecomment-346086049) that wasn't entirely correct.

taquitos:
The reason why I think `^` works is because you pointed out that we read the whole file in as a single line, and not as multi-line.
So, `^` is effectively the same as `\A` in this particular case because the string is just a single line.

revolter:
I see, I'm not accustomed to Ruby regex, so I just use my js/py regex knowledge.

But I think that in this case, the string isn't a single line, it doesn't remove the `\r\n`s, it still is a big string consisting of multiple lines of text (text separated by a newline character).

revolter:
@janpio, Could you test the `^` variant and `\A` like you did in your first comment?

taquitos:
Good point... we can test in `irb` by loading up one of these newly generated files and adding stuff to the start 😄

revolter:
```
irb(main):001:0> string = "something here
irb(main):002:0"
irb(main):003:0" <!-- Thanks for contributing to _fastlane_! Before you submit your pull request, please make sure to check the following boxes by putting an x in the [ ] -->"
=> "something here\n\n<!-- Thanks for contributing to _fastlane_! Before you submit your pull request, please make sure to check the following boxes by putting an x in the [ ] -->"
irb(main):004:0> string
=> "something here\n\n<!-- Thanks for contributing to _fastlane_! Before you submit your pull request, please make sure to check the following boxes by putting an x in the [ ] -->"
irb(main):005:0> string =~ /^(#|<!--)/ ? "yes" : "no"
=> "yes"
irb(main):006:0> string =~ /\A(#|<!--)/ ? "yes" : "no"
=> "no"
```

So `\A` is the better choice to ensure no text is added before the header.

taquitos:
Ahhh, nice!

PR #466, opened 2017-11-21
add more docs on environment variables

janpio:
As not everybody knows how to set them (me included!), I googled and found a comment of @krausefx that greatly explained it:

https://github.com/fastlane/fastlane/issues/319#issuecomment-105842100

Adapted these and added them to advanced.md


taquitos:
Yes, this is great, I know I've had questions in the past around this 🎈:goat:

PR #470, opened 2017-11-22
Mention locale environment variables in the docs

milch:
See fastlane/fastlane#10988 

This makes locale environment variables that need to be set more explicit by mentioning them during setup, in the FAQs and in the CI setup guide

### DIFF
--- a/docs/getting-started/android/setup.md
+++ b/docs/getting-started/android/setup.md
@@ -24,7 +24,7 @@
 
 ### Set up environment variables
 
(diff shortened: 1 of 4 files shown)


docs/getting-started/android/setup.md
```diff
@@ -22,6 +22,17 @@
 </tr>
 </table>
 
+### Set up environment variables
+
+fastlane requires some environment variables set up to run correctly. In particular, having your locale not set to a UTF-8 locale will cause issues with building and uploading your build. In your shell profile add the following lines:
```
revolter: Maybe use `_fastlane_` as in the other places?

milch: Good point, thanks 👍

PR #472, opened 2017-11-23
Update tab auto complete FAQ

revolter:
This adds fish to the list of supported shells for tab auto complete.
This enables searching for "auto complete" to find this FAQ.

The problem with this is that previous link (https://docs.fastlane.tools/faqs/#enable-bashzsh-tab-completion-for-fastlane-lane-names) wouldn't work anymore.

Is this something we can live with?

### DIFF
--- a/docs/faqs.md
+++ b/docs/faqs.md
@@ -102,13 +102,15 @@ Set the `FASTLANE_DISABLE_COLORS` environment variable to disable ANSI colors (e
 export FASTLANE_DISABLE_COLORS=1
 ```
 
-### Enable _bash_/_zsh_/_fish_ tab auto complete for fastlane lane names
+### Enable tab auto complete for fastlane lane names
+
+Supported shells: _bash_, _zsh_, _fish_.
 
 ```no-highlight
 fastlane enable_auto_complete
 ```
 
-Follow the on screen prompt to add a line to your _bash_/_zsh_ profile.
+Follow the on screen prompt to add a line to your _bash_/_zsh_/_fish_ profile.
 
 ### "User interaction is not allowed" when using _fastlane_ via SSH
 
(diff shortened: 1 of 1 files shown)


janpio:
I remember that with Jekyll it was possible to add "empty" anchors, so one could make sure old anchor links still worked. Maybe the same is possible with mkdocs?

revolter:
Would be great if

```
<!-- ### Enable _bash_/_zsh_ tab completion for fastlane lane names -->
### Enable _bash_/_zsh_/_fish_ tab auto complete for fastlane lane names
```

would be rendered to:

```
<!-- <h3 id="enable-bashzsh-tab-completion-for-fastlane-lane-names">
    <a name="user-content-enable-bashzsh-tab-completion-for-fastlane-lane-names" href="#enable-bashzsh-tab-completion-for-fastlane-lane-names" class="headeranchor-link" aria-hidden="true">
        <span class="headeranchor"></span>
    </a>
    Enable <em>bash</em>/<em>zsh</em> tab completion for fastlane lane names
</h3> -->

<h3 id="enable-bashzshfish-tab-auto-complete-for-fastlane-lane-names">
    <a name="user-content-enable-bashzshfish-tab-auto-complete-for-fastlane-lane-names" href="#enable-bashzshfish-tab-auto-complete-for-fastlane-lane-names" class="headeranchor-link" aria-hidden="true">
        <span class="headeranchor"></span>
    </a>
    Enable <em>bash</em>/<em>zsh</em>/<em>fish</em> tab auto complete for fastlane lane names
</h3>
```

janpio:
I don't think that would work, uncommented code is not interpreted at all I think.

Can you maybe somehow add custom HTML? Then a `<a id='enable-bashzsh-tab-completion-for-fastlane-lane-names' />` should be enough I think.

joshdholtz:
I think the best thing to do would be to rename the heading and do something like...
```
### Enable tab auto complete for fastlane lane names
Supported shells - _bash_, _zsh_, _fish_
```

I don't _think_ we need to worry about supporting a broken link for this example? It will still bring the user to this page and now this link will work any future additions/removals of shells

PR #477, opened 2017-11-29
Advanced: new paragraph "Lane Properties"

janpio:
... that completes the "Lane Context" paragraph and moves "Lane" over there, but also adds "Platform" and "Default Platform".


docs/advanced.md
```diff
@@ -314,6 +314,23 @@ Although it kind of defeats the purpose of using them in the first place (not to
 ENV["DELIVER_USER"] = "felix@krausefx.com"
 ```
 
+# Lane Properties
+
+It can be useful to dynamically access properties of the current lane. These are available in `lane_context`
```
revolter: This should end with a `:`.

docs/advanced.md
```diff
@@ -314,6 +314,23 @@ Although it kind of defeats the purpose of using them in the first place (not to
...
+
+```ruby
+lane_context[SharedValues::PLATFORM_NAME]        # Platform name, e.g. `ios`, `android` or empty (for root level lanes)
+lane_context[SharedValues::LANE_NAME]            # The name of the current lane (stays the same when switching lanes, includes platform name)
+lane_context[SharedValues::DEFAULT_PLATFORM]     # Default platform
+```
+
+and environment variables
```
revolter: Same as the previous comment.

docs/advanced.md
```diff
@@ -314,6 +314,23 @@ Although it kind of defeats the purpose of using them in the first place (not to
 ENV["DELIVER_USER"] = "felix@krausefx.com"
 ```
 
+# Lane Properties
+
+It can be useful to dynamically access properties of the current lane. These are available in `lane_context`
+
+```ruby
+lane_context[SharedValues::PLATFORM_NAME]        # Platform name, e.g. `ios`, `android` or empty (for root level lanes)
+lane_context[SharedValues::LANE_NAME]            # The name of the current lane (stays the same when switching lanes, includes platform name)
```
revolter: I'd personally change it to `, applies to platform name too` or even better, move it to the previous line so even if it'd be duplicated, it would be clearer.

revolter:
How can you find out all the available ENV variables?

docs/advanced.md
janpio: Hm, "includes platform name" is not meant as a qualifier to "stays the same...". The lane name actually includes the platform name: `ios foo` for lane `foo` for platform `ios`.

janpio:
No idea. I found these when I looked through the code for the other stuff.
There are loads more, but these have the same content so it seems useful to mention them.

docs/advanced.md
janpio: (not actually outdated @Github...)

revolter: Ohhh, I thought that the platform stays the same. Maybe

> The name of the current lane preceded by the platform name.

? How does it sound to you?

janpio: Good.

PR #479, opened 2017-11-30
Shell-related updates

jdee:
Removed mention of `other_action.sh` in favor of `Actions.sh`, which is available in all actions simply as `sh`, without a directory change.

Added discussion of shell escaping with `sh` (not yet released).

Some other corrections and improvements to the shell-out section.


docs/advanced.md
```diff
@@ -1018,7 +1018,7 @@ UI.error "Something unexpected happened in my_new_action. Attempting to continue
 |verbose|Displays verbose output in the default output color|
 |header|Displays a message in a box for emphasis|
 
-Methods ending in and exclamation point (`!`) terminate execution of the current
+Methods ending in an exclamation point (`!`) terminate execution of the current
```
revolter: Shouldn't this be `exclamation mark`?

docs/advanced.md
```diff
@@ -1105,27 +1106,28 @@ global variable will indicate the completion status of the command.
 
 If the command to be executed is not found, `Errno::ENOENT` is raised.
 
-### Using the sh action
+### Using the sh method
 
-You can also use the built-in `sh` Fastlane action:
+You can also use the built-in `Actions.sh` method:
```
janpio: Why call is "`Actions.sh` method" if I use it by just calling `sh`?

jdee: For one thing, this introduces the method and so provides more information about where it comes from. (It took me a while to track down where all this is happening.) It happens to be available in any class method in the `Actions` namespace, so within an action (including in a plugin) it can just be called as `sh`. If you're in a helper, e.g., you cannot just call `sh`. It has to be `Actions.sh` (or even `Fastlane::Actions.sh`). I think it's valuable to mention `Actions.sh` once.

docs/advanced.md
jdee: I think "exclamation point" and "exclamation mark" are synonymous. This is more natural to me, but I think either one is acceptable.

docs/advanced.md
revolter: Makes me think of a shell script named `Actions` that has the standard `sh` extension.

docs/advanced.md
revolter: I'm not a native english speaker and I only knew about "exclamation mark", so then disregard this review.

PR #496, opened 2017-12-14
Update index.md

artrmz:
Lane shouldn't be called 'appstore' as it's action's name.


artrmz:
I  signed it!

KrauseFx:
Ahhh, yesss, nice catch @artrmz, thanks for the PR. I just did a quick search and it looks like we have that lane name across multiple pages on the docs repo, would you mind replacing all of them? 

Also, instead of `sample_appstore` could we use `lane :release`? Sounds a little nicer :)

revolter:
Most probably [there ones](https://sourcegraph.com/search?q=%5C:appstore&sq=repo:%5Egithub%5C.com/fastlane/docs%24).

PR #501, opened 2017-12-15
Put the installer script method first

tbodt:
Installing via Homebrew gives an inferior user experience, with constant prompting to use a Gemfile and an associated 2-second startup delay.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -59,8 +59,8 @@ xcode-select --install
 
 <table width="100%">
 <tr>
-<th width="33%"><a href="http://brew.sh">Homebrew</a></td>
 <th width="33%">Installer Script</td>
+<th width="33%"><a href="http://brew.sh">Homebrew</a></td>
 <th width="33%">Rubygems</td>
 </tr>
 <tr>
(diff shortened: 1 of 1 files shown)


janpio:
I saw a PR about that 2 sec thing, what else do you thing makes an "inferior user experience" for the Homebrew install? (Honestly curious)

tbodt:
It's really just the 2 second wait. I don't care enough about slow startup times to be bothered to create a Gemfile.

KrauseFx:
Thanks for the PR, I agree the homebrew installation isn't perfect. I feel like the Gemfile instructions should probably be the number one choice for installation, what do y'all think?

janpio:
2 second wait is gone: https://github.com/fastlane/fastlane/pull/11240/files

tbodt:
@KrauseFx My experience is it's really, really annoying to have to put `bundle exec` in front of everything you do.

revolter:
You can create a `fl` file in the project's root containing:

```
#!/usr/bin/env bash

bundle exec fastlane $@
```

give it executable permissions and use it like `./fl lane_name`.

tbodt:
But that requires creating the file for all my projects. I’d much rather just type `fastlane` and be done.

~Theodore

revolter:
You can also create an alias for it:

```
alias fl="bundle exec fastlane $@"
```

but then every team member has to add this.

revolter:
Hey @tbodt, Sorry for the long wait. The install instructions were actually completely re-done recently: https://docs.fastlane.tools/#install-fastlane. Is this current version better?

KrauseFx:
Closing this now, thanks everyone 👍

PR #502, opened 2017-12-18
Replace HTML "Installation" table with markdown table, change orientation from horizontal to vertical

janpio:
- Replaces the HTML table with a Markdown table
- Changes orientation of table from horizontal to vertical

Continuation from https://github.com/fastlane/fastlane/pull/11198

### DIFF
--- a/docs/getting-started/android/setup.md
+++ b/docs/getting-started/android/setup.md
@@ -4,23 +4,11 @@
 
 ### Choose your installation method:
 
-<table width="100%">
-<tr>
-<th width="33%"><a href="http://brew.sh">Homebrew</a></td>
-<th width="33%">Installer Script</td>
-<th width="33%">Rubygems</td>
-</tr>
-<tr>
-<td width="33%" align="center">macOS</td>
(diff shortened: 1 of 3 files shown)


revolter:
```
| Method | OS support | Description |
| --- | --- | --- |
| [Homebrew](http://brew.sh) | macOS | `brew cask install fastlane`                                                                                                          |
| Installer Script | macOS | [Download the zip file](https://download.fastlane.tools). Then double click on the `install` script (or run it in a terminal window). |
| RubyGems | macOS or Linux with Ruby 2.0.0 or above | `sudo gem install fastlane -NV` |
```

also works, as [spaces are optional and you only need at least 3 dashes for the header](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables).

janpio:
Don't you like your markdown to be pretty as well?

revolter:
Well, it depends. Looking at the diff here on GitHub is hard, even on a 27" display, but it's not that important, just wanted to let you know. I'd leave it that way so it looks better when editing it locally. So it's your call.

PR #512, opened 2018-01-03
Add information about incrementing the build number

KrauseFx:


docs/getting-started/ios/appstore-deployment.md
```diff
@@ -76,4 +76,23 @@ If you don't have any push certificates already, _get_push_certificate_ will cre
...
 
+<details>
+<summary>Incrementing the build number</summary>
+
+The code sample below will use the latest build number from iTunes Connect and temporarily set it. 
+
+```ruby
+lane :beta do
+  increment_build_number(
+    build_number: app_store_build_number + 1,
```
janpio: Where does `app_store_build_number` come from?

KrauseFx: It's here https://docs.fastlane.tools/actions/app_store_build_number/

or what do you mean? The testflight action is just an alias

janpio: Ah, this is a method that is being called, not a variable that contains some value. 
Confusing or did I miss anything?

KrauseFx: Yep, it's an action

KrauseFx: What do you propose? multiple lines?

janpio: Probably nothing, my PHP brain just wants to see something indicating a function call here: `app_store_build_number()`. But if this is the way it's done in ruby, that's fine - I just didn't get it on first look.

PR #514, opened 2018-01-06
Update iOS setup docs to be better

KrauseFx:


docs/getting-started/ios/setup.md
```diff
@@ -38,15 +27,13 @@ fastlane init
...
+```no-highlight
 fastlane init swift
 ```
 
 **Swift setup is still in beta. See [Fastlane.swift docs](https://docs.fastlane.tools/getting-started/ios/fastlane-swift/) for more information.**
 
-You'll be asked for your Apple ID, and _fastlane_ will automatically generate a fastlane configuration for you, based on the information provided.
-
-You can see the newly created _fastlane_ directory, with all your existing metadata fetched from iTunes Connect.
+Depending on what kind of setup you choose, different files will be set up for you. If you chose to download the existing app metadata, you'll end up new folders that look like this:
```
lacostej: Maybe

```
Depending on how you setup your _fastlane_ project, different files will be created.
```
is better for the first sentence?

And "you'll end up WITH new folders..." or even "you'll end up with a directory structure that looks like this:" ?

KrauseFx: With the new `fastlane init`, you can actually choose what you want to use fastlane for (https://github.com/fastlane/fastlane/pull/11386), so I think this should be ok

KrauseFx:
Is it ok to merge for now?

docs/getting-started/ios/setup.md
```diff
@@ -56,9 +43,20 @@ The most interesting file is `fastlane/Fastfile`, which contains all the informa
 
 _fastlane_ created all the required files for you, now you can go ahead and customise it to generate screenshots or to automatically distribute new builds
 
-- [_fastlane_ screenshots for iOS and tvOS](screenshots.md)
-- [iOS Beta deployment using _fastlane_](beta-deployment.md)
-- [iOS App Store deployment using _fastlane_](appstore-deployment.md)
+- [Generate localized iOS screenshots for the App Store](screenshots.md)
```
janpio: no more tvOS?

KrauseFx: We don't officially support tvOS currently 😢

PR #525, opened 2018-01-16
Refactor installation methods for fastlane

KrauseFx:
This is not done yet, but even if we automatically generate a Gemfile for the user, we still need to install fastlane for them.
1) to install fastlane and run `fastlane init`
2) to install `bundler` for them
See implementation https://github.com/fastlane/fastlane/pull/11582

So once it's installed, we'd then automatically setup and use a Gemfile for the user, and even with a proposal to automatically do `bundle exec` for the user with https://github.com/fastlane/fastlane/pull/11588

screen shot 2018-01-16 at 3 24 03 pm
screen shot 2018-01-16 at 3 23 51 pm
screen shot 2018-01-16 at 3 23 36 pm

Let me know what y'all think 👍

### DIFF
--- a/docs/getting-started/ios/setup.md
+++ b/docs/getting-started/ios/setup.md
@@ -7,13 +7,17 @@ Make sure you have the latest version of the Xcode command line tools installed:
 xcode-select --install
 ```
 
-### Choose your installation method:
+Install _fastlane_ using 
+```no-highlight
+[sudo] gem install fastlane -NV
+```
+or alternatively using `brew cask install fastlane`
 
-| Method                     | OS support                              | Description                                                                                                                           |
-|----------------------------|-----------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------|
-| [Homebrew](http://brew.sh) | macOS                                   | `brew cask install fastlane`                                                                                                          |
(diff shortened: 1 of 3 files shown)


taquitos:
❤️:goat:

KrauseFx:
Not merging yet, this need to be applied to all other pages also, including the _fastlane_ code that generates the READMEs for match

docs/getting-started/ios/setup.md
```diff
@@ -7,13 +7,17 @@ Make sure you have the latest version of the Xcode command line tools installed:
...
 ```
 
-### Choose your installation method:
+Install _fastlane_ using 
+```no-highlight
+[sudo] gem install fastlane -NV
+```
+or alternatively using `brew cask install fastlane`
+
+Navigate to your iOS or Android app and run
```
janpio: "iOS or Android app" => "project directory/folder" maybe?

janpio:
Definitely like the simplicity.

(Note that this might get more complicated again with Windows and Linux support by the way - but we will tackle this when this is done. For now making it the best for macOS is the right thing to do.)

docs/getting-started/ios/setup.md
KrauseFx: Nice, thanks

KrauseFx:
Updating now

KrauseFx:
Alright, this is ready now 👍

PR #530, opened 2018-01-23
Update Setup Instructions with New Service Role

pburtchaell:
As per the discussion on https://github.com/fastlane/fastlane/issues/11565, the Service Account Actor role is depreciated. This PR updates the documentation with the name of the new role you should use.


pburtchaell:
I signed it!

pburtchaell:
Looks like there's also the same error here: https://docs.fastlane.tools/actions/supply/

We should probably create another PR to patch that one too.

KrauseFx:
Thanks 👍

PR #539, opened 2018-02-07
Update SSL FAQ to include Hockey CDN bypass

ZarekParker:
I have encountered SSL errors with Hockey a few times, resolved only by bypassing the CDN.  This doesn't appear anywhere in the SSL FAQs and could provide help to others.

### DIFF
--- a/docs/faqs.md
+++ b/docs/faqs.md
@@ -20,6 +20,17 @@ If you use `rvm`, try the following
 rvm osx-ssl-certs update all
 ```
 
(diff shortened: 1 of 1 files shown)


ZarekParker:
I signed it!

KrauseFx:
Thanks for your contribution @ZarekParker. I think this should actually be part of the `hockey` docs, which you can edit over here: https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/hockey.rb#L352

Thanks 👍

PR #542, opened 2018-02-15
Add Match manual for handling multiple apps per developer account

vmalyi:
A clear guidance on how to use the same development and/or distribution certificate for multiple apps from the same development team.

### DIFF
--- a/docs/actions/match.md
+++ b/docs/actions/match.md
@@ -163,7 +163,7 @@ Then all your team has to do is `fastlane certificates` and keys, certs and prof
 
 #### Handle multiple apps per developer/distribution certificate
 
-If you want to use single developer and distribution certificate for multiple apps belonging to the same development team, you may use the same signing identities repo and branch to store signing identities for your apps:
+If you want to use a single developer and distribution certificate for multiple apps belonging to the same development team, you may use the same signing identities repository and branch to store the signing identities for your apps:
 
 Matchfile for App #1:
 ```no-highlight
@@ -176,7 +176,7 @@ Matchfile for App #2:
 git_url "https://github.com/example/example-repo.git"
 git_branch "master"
 ```
-Match will reuse developer/distribution certificate and will create separate provisioning profiles for each app.
+_match_ will reuse certificates and will create separate provisioning profiles for each app.
 
 #### Passphrase
 
--- a/docs/actions/match.md
+++ b/docs/actions/match.md
@@ -163,7 +163,7 @@ Then all your team has to do is `fastlane certificates` and keys, certs and prof
 
 #### Handle multiple apps per developer/distribution certificate
 
-If you want to use a single developer and distribution certificate for multiple apps belonging to the same development team, you may use the same signing identities repository and branch to store the signing identities for your apps:
+If you want to use a single developer and/or distribution certificate for multiple apps belonging to the same development team, you may use the same signing identities repository and branch to store the signing identities for your apps:
 
 Matchfile for App #1:
 ```no-highlight
--- a/docs/actions/match.md
+++ b/docs/actions/match.md
@@ -161,6 +161,23 @@ end
 
 Then all your team has to do is `fastlane certificates` and keys, certs and profiles for all targets will be synced.
 
+#### Handle multiple apps per developer/distribution certificate
+
+If you want to use single developer and distribution certificate for multiple apps belonging to the same development team, you may use the same signing identities repo and branch to store signing identities for your apps:
+
+Matchfile for App #1:
+```no-highlight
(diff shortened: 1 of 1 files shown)


docs/actions/match.md
```diff
@@ -161,6 +161,23 @@ end
...
+git_url "https://github.com/example/example-repo.git"
+git_branch "master"
+```
+
+Matchfile for App #2:
+```no-highlight
+git_url "https://github.com/example/example-repo.git"
+git_branch "master"
+```
+Match will reuse developer/distribution certificate and will create separate provisioning profiles for each app.
```
janpio: Match => _match_ (with underscores for italic)

janpio: ... reuse developer/distrivution certificate ... => ... resue the certificate ...

docs/actions/match.md
```diff
@@ -161,6 +161,23 @@ end
 
 Then all your team has to do is `fastlane certificates` and keys, certs and profiles for all targets will be synced.
 
+#### Handle multiple apps per developer/distribution certificate
+
+If you want to use single developer and distribution certificate for multiple apps belonging to the same development team, you may use the same signing identities repo and branch to store signing identities for your apps:
```
janpio: ... use single ... => ... use a single ...
repo => repository
... store signing ... => ... store the signing ...

vmalyi: Done

docs/actions/match.md
vmalyi: Done

docs/actions/match.md
```diff
@@ -161,6 +161,23 @@ end
 
 Then all your team has to do is `fastlane certificates` and keys, certs and profiles for all targets will be synced.
 
+#### Handle multiple apps per developer/distribution certificate
+
+If you want to use a single developer and distribution certificate for multiple apps belonging to the same development team, you may use the same signing identities repository and branch to store the signing identities for your apps:
```
janpio: ... developer and distribution ...: This should be "or" probably?

vmalyi: @janpio both "or" and "and" are valid in this case since the user can have a single developer certificate, a single distribution certificate which can be used for multiple apps. Each app, in this case, will become individual provisioning profiles based on certificates from above.

janpio: Yep, that is the perfect solution. Actually makes it clearer what is possible! 👍

vmalyi:
Added another correction as requested

janpio:
🥇  Thanks!

docs/actions/match.md
```diff
@@ -161,6 +161,23 @@ end
...
 Then all your team has to do is `fastlane certificates` and keys, certs and profiles for all targets will be synced.
 
+#### Handle multiple apps per developer/distribution certificate
+
+If you want to use a single developer and/or distribution certificate for multiple apps belonging to the same development team, you may use the same signing identities repository and branch to store the signing identities for your apps:
+
+Matchfile for App #1:
+```no-highlight
+git_url "https://github.com/example/example-repo.git"
+git_branch "master"
```
revolter: These should use parens, since we recently decided with that style instead:

```
git_url("https://github.com/example/example-repo.git")
git_branch("master")
```

docs/actions/match.md
```diff
@@ -161,6 +161,23 @@ end
 
 Then all your team has to do is `fastlane certificates` and keys, certs and profiles for all targets will be synced.
 
+#### Handle multiple apps per developer/distribution certificate
+
+If you want to use a single developer and/or distribution certificate for multiple apps belonging to the same development team, you may use the same signing identities repository and branch to store the signing identities for your apps:
+
+Matchfile for App #1:
+```no-highlight
```
revolter: This is actually `ruby`

joshdholtz:
@basdrm This is actually an autogenerated file 🙃 (which you can see at the top)
```
<!--
 This file is auto-generated and will be re-generated every time the docs are updated.
 To modify it, go to its source at https://github.com/fastlane/fastlane.
 -->
```

You will want to go change the documentation over on the main fastlane repo instead - https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/sync_code_signing.md

Sorry for the confusion!

joshdholtz:
@vmalyi Do you want to handle this? Or would you like me to take care of this for you? 😊

janpio:
See https://github.com/fastlane/fastlane/pull/13003

docs/actions/match.md
janpio: Turns out it is not: https://github.com/fastlane/fastlane/pull/13102

PR #553, opened 2018-03-05
Added deliver documentation for all allowed metadata folder options

joshdholtz:
Related to https://github.com/fastlane/fastlane/issues/11988

| Description | Screenshot |
|---|---|
| All | screen shot 2018-03-05 at 7 13 04 am|
| Non-Localized | screen shot 2018-03-05 at 7 13 23 am|
| Localized | screen shot 2018-03-05 at 7 13 23 am|
| Trade Representative | screen shot 2018-03-05 at 7 13 31 am|
| Review | screen shot 2018-03-05 at 7 13 42 am|

### DIFF
--- a/docs/actions/deliver.md
+++ b/docs/actions/deliver.md
@@ -427,6 +427,86 @@ _deliver_ uses the following techniques under the hood:
 no, en-US, en-CA, fi, ru, zh-Hans, nl-NL, zh-Hant, en-AU, id, de-DE, sv, ko, ms, pt-BR, el, es-ES, it, fr-CA, es-MX, pt-PT, vi, th, ja, fr-FR, da, tr, en-GB
 ```
 
+## Available Metadata Folder Options
+
+_deliver_ allows for metadata to be set through `.txt` files in the metadata folder. This metadata folder location is defaulted to `./fastlane/metadata` but can be overridden through the `metadata_path` parameter. Below are all allowed metadata options.
+
+### Non-Localized Metadata
+
+<details>
+<summary>View all</summary>
+
+Key | Editable While Live | Directory | Filename
+----|--------|--------|--------
+  `copyright` | No | `<metadata_path>` | `copyright.txt`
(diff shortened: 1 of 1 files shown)


revolter:
I don't know how `deliver` works behind the scenes, but, could these be generated automatically? So they don't get out of sync in the future.

janpio:
Do you think using "View all" is absolutely necessary? Problem with the `<details>` tag is that it can be confusing if someone finds a page via Google, then can not find the actual text that popped up on the page directly (or when scanning over the content of the page). I think this is important enough to just have the tables in there.

joshdholtz:
@revolter These _could_ be generated automatically since all of these value are stored in constants in `upload_metadata.rb`. I haven't done any of that before with these docs so I don't really know much work it is to implement that but I can totally look into 💪 

@janpio I don't think the "View all" is necessary but the tables do take up a lot of height. I'd be totally down for removing the view all, though 😇 I think its information that is important enough to always be shown. Thoughts?

revolter:
It would be great if searching for some text would find out even if the details are collapsed, but sadly, it's not the case currently. So yes, removing the details would probably be a good idea.

Also, this really needs to be programmatically generated, as it will be overwritten. Notice the comment at the start of the file:

> This file is auto-generated and will be re-generated every time the docs are updated.
> To modify it, go to its source at https://github.com/fastlane/fastlane.

revolter:
Oh, you could also make these exact changes on this file instead: https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/upload_to_app_store.md

joshdholtz:
@revolter OMG, I didn't even see that 😱

revolter:
@joshdholtz, I added that 😛

joshdholtz:
@revolter Well, I guess I close this an open a PR in https://github.com/fastlane/fastlane then? 🙃

janpio:
+1 on all the things you both said. Good work going on here :)

PR #557, opened 2018-03-09
Service Account User now under 'Service Accounts'

luciolucio:
Service Account User moved to the Service Accounts menu item. I was just setting up a new Android app today and noticed this.

screenshot at mar 09 18-36-47


luciolucio:
That commit did not have my email address in it for some reason. I'll fix that and re-send.

PR #570, opened 2018-03-26
Update Fastlane docs to reflect current best practices for CircleCI

zzak:
* Since we specify working_directory store_artifacts and store_test_results can be relative to that
* You no longer need to cp the report.junit as I've fixed our test results parser to handle `.junit` files

---

**[Recent build showing latest demo config](https://circleci.com/gh/CircleCI-Public/circleci-demo-ios/21#config/containers/0)**

You can also view the full config [here](https://github.com/CircleCI-Public/circleci-demo-ios/blob/master/.circleci/config.yml).

screenshot from 2018-03-26 19-14-52

### DIFF
--- a/docs/best-practices/continuous-integration.md
+++ b/docs/best-practices/continuous-integration.md
@@ -206,7 +206,7 @@ jobs:
       xcode: "9.0"
     working_directory: /Users/distiller/project
     environment:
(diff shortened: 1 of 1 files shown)


zzak:
I signed it!

PR #574, opened 2018-04-04
Updated numeric bullet spacing

chefnobody:
Makes it easier to read the code block above step 12, and step 12 itself. This is under Getting Started and not the Actions, per-se. Still looking to see if this needs to be a change made in source. Any tips?


chefnobody:
I signed it!

Issue #575, opened 2018-04-04
Incorrect indented code formatting


revolter:
Having code block indented with 4 spaces (like in a list) renders incorrectly:

![screen shot 2018-04-04 at 21 55 09](https://user-images.githubusercontent.com/5748627/38328180-dd476254-3852-11e8-93e1-493b5b796e3c.png)
_https://docs.fastlane.tools/getting-started/ios/screenshots/#setting-up-snapshot_

The issue is caused by `Python-Markdown` and was reported here: https://github.com/Python-Markdown/markdown/issues/53, which is on the [roadmap for 3.0](https://github.com/Python-Markdown/markdown/wiki/3.0-Roadmap).

We can either wait for it to be fixed, then released, then a release from `mkdocs` including it, or we can remove the indentation and change the next list element to `12.` instead of `1.`.

joshdholtz:
😱 Ummmm... _sigh_ I guess make it 12? 🤔

janpio:
Yep, I think I "reworked" that document and seems only semi checked if it really rendered on the site, probably only checked on Github itself. Possible it even was that way before :/

revolter:
GitHub renders it different from mkdocs.

janpio:
Is this fixed now @revolter?

revolter:
No, we're still waiting for [markdown](https://github.com/Python-Markdown/markdown/releases) to release version `3.0`.

janpio:
Current version of `markdown` now is `3.1.1`. What do we have to do to get this used and make the problem go away @revolter?

revolter:
From what I see, that issue is still opened, and was removed from the 3.0 roadmap, and it's also not checked it in the actual [roadmap](https://github.com/Python-Markdown/markdown/wiki/3.0-Roadmap), so maybe it was postponed. But I didn't look too much into it.

PR #577, opened 2018-04-07
Temporary fix incorrect indented code formatting

revolter:
Related to #575


janpio:
(Ugh - but if it works...)

revolter:
Not much we can do for now :(

Issue #560, opened 2018-03-12
Replace pip with pipenv


revolter:
I would want to know others opinion on replacing pip with pipenv.

I find it way better and easier to use. You don't have to create/modify either the `requirements` file either the `requirements-to-freeze` file, nor load it when installing the dependencies.

The only steps are:
- `pipenv --three` to create a virtual env
- `pipenv shell` to load the virtual env
- `pipenv install` to install the dependencies in the virtual env

And `pipenv install <dependency>` to add a new dependency (currently 3 steps).

Also, the current README doesn't mention virtual envs, so following the steps in it causes the packages to be installed globally.

I can do the migration and README/docs update myself.

joshdholtz:
>Pipenv — the officially recommended Python packaging tool from Python.org, free (as in freedom).

Seems legit 😊 The current toolset "works" for me but I am not a Python pro so I don't really have any opinion for or against this.

revolter:
It's actually just a wrapper over `virtualenv`.

janpio:
Can this be closed @revolter?

PR #579, opened 2018-04-07
Update to python3 and add pipenv

revolter:
As I mentioned in #560, this simplifies the installation instructions and adding a new dependency.

Also, the current `requirements.txt` and `requirements-to-freeze.txt` are incorrect for two reasons:
- `markdown-include` is missing from the `requirements-to-freeze.txt`
- some packages like `backports-abc` don't seem to be needed, as they're missing from the auto-generated `Pipfile.lock`

This way, you can also make sure you use a specific version (or a minimum), more similar to `Gemfile` and `Gemfile.lock` than the `requirements*.txt` files.

It's also a wrapper over `virtualenv` and `pip`, so it's not actually replacing anything.

I would also want to know if updating to Python 3 has any drawbacks, even though CircleCI succeeded.


minuscorp:
Sincerely, I never liked how Python made the dependency management. IMO pip is a tool where the concept: "works in my computer" gets real. I'm happy to see new alternatives and seems pretty solid for me.

revolter:
Yup, pipenv is the best. I also love that it manages the virtual environments automatically, and also stores them outside the project's directory.

circle.yml
```diff
@@ -1,11 +1,12 @@
...
   python:
-    version: 2.7.11
+    version: 3.6.0
   ruby:
     version: 2.3.0
 dependencies:
   pre:
-    - pip install -r requirements.txt
+    - pip install pipenv
+    - pipenv install
```
revolter: Can't get easier than this ❤️

getaaron:
This looks great 👍

PR #604, opened 2018-05-08
Update running-tests.md

0xPr0xy:
Remove `run_tests` action documentation from the android testing documentation since this is confusing.

### DIFF
--- a/docs/getting-started/android/running-tests.md
+++ b/docs/getting-started/android/running-tests.md
@@ -63,5 +63,6 @@ error do |ex|
 end
 ```
 
-The above example uses Hipchat, but _fastlane_ supports [many other services out there](/actions#notifications). 
+#### Other services
 
+The above example uses Hipchat, but _fastlane_ supports [many other services out there](/actions#notifications). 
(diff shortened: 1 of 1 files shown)


0xPr0xy:
I signed it!

revolter:
Thank you for the Pull Request 😃 Could you please elaborate a bit on the confusing part? Could we change/reword it so it gets clearer, instead of removing it?

0xPr0xy:
We could rewrite it to make it more clear.
However, we would do this by clarifying this action is only available for the iOS platform.
This is not the best solution in my opinion since the user is browsing the testing documentation page for the Android platform. Therefore it makes more sense to have it on the iOS testing page and not on the Android testing page.

Hope this clarifies my intent.
Kind regards, 0x

revolter:
Oh, I see, I didn't know that `run_tests` is an iOS only action.

joshdholtz:
@0xPr0xy Are you cool if we merge this PR with @revolter changes in it? Need to confirm before we merge 😊

0xPr0xy:
Definately! 👍

joshdholtz:
Everyone has signed CLA and @0xPr0xy has given permission so merging 👍

PR #612, opened 2018-05-23
Fixed typo (recommend way -> recommended way)

simonbromberg:
Previously: "More information about the recommend way to install gems"


simonbromberg:
I signed it

PR #620, opened 2018-06-07
Remove outdated Apple developer news URL

josh-:
It appears that Apple only persists the last year's worth of articles on `developer.apple.com/news/`, and browsing to an article with an `id` older than that just goes to the main news list.

This PR updates the link in `screenshots.md` to use a permanent <archive.org> link.

Thanks team! ✨

### DIFF
--- a/docs/getting-started/ios/screenshots.md
+++ b/docs/getting-started/ios/screenshots.md
@@ -25,7 +25,7 @@ The positive side of this approach is that your screenshots will all be crisp an
 
 ### Scaled Screenshots via iTunes Connect
 
-With the [iTunes Connect screenshot update](https://developer.apple.com/news/?id=08082016a) from August 2016, you can now use one set of screenshots for all available devices and screenshots. iTunes Connect will automatically scale the images for you, so that each device renders the same image.
+iTunes Connect allows you to use one set of screenshots per device type, which will then be scaled to the appropriate size when viewed in the App Store on a user's device.
 
 While this is convenient, this approach has the same problems as the device frame approach: The screenshots don't actually show how the app looks on the user's device. It's a valid way to start though, since you can gradually overwrite screenshots for specific languages and devices.
 


docs/getting-started/ios/screenshots.md
```diff
@@ -25,7 +25,7 @@ The positive side of this approach is that your screenshots will all be crisp an
 
 ### Scaled Screenshots via iTunes Connect
 
-With the [iTunes Connect screenshot update](https://developer.apple.com/news/?id=08082016a) from August 2016, you can now use one set of screenshots for all available devices and screenshots. iTunes Connect will automatically scale the images for you, so that each device renders the same image.
+With the [iTunes Connect screenshot update](http://web.archive.org/web/20161118092153/https://developer.apple.com/news/?id=08082016a) from August 2016, you can now use one set of screenshots for all available devices and screenshots. iTunes Connect will automatically scale the images for you, so that each device renders the same image.
```
revolter: Is there a problem if we use the latest snapshot: `https://web.archive.org/web/20170728111611/https://developer.apple.com/news/?id=08082016a`?

janpio: Does it make a difference?

KrauseFx: Linking to web archive isn't super cool tbh, is the news available anywhere else still, so we can link directly? 

Also, since it's been about 2 years since then, we probably don't need to link at all, right?

josh-: Good point @KrauseFx, the most up-to-date information I could find is [this App Store Connect help page](https://help.apple.com/app-store-connect/#/devd1093d90d). So I could reword the sentence to something like this?

> [App Store Connect](https://help.apple.com/app-store-connect/#/devd1093d90d) allows you to upload one set of screenshots per device, which will then be scaled to the appropriate size when viewed in the App Store on a user's device.

josh-: (That could also be written as 'iTunes Connect' – bit challenging to know which one to use in this period of renaming)

KrauseFx: Yeah let's just not link and simplify that text, since it's been so long ago 👍

josh-: Sounds good @KrauseFx, just updated the commit! 😄

KrauseFx:
Beautiful, thanks everyone 🎉

PR #639, opened 2018-07-30
remove duplicated fastlane init

janpio:
Noticed `fastlane init` was mentioned twice, removed once.


janpio:
Nice to be back :)

Ohoh, the netlify thingie doesn't like the change. Huh?

janpio:
Any idea why it failed? Just flaky?

KrauseFx:
I don't know actually

PR #658, opened 2018-08-01
Highlight "Actions" in navigation when on action docs

### LINKED ISSUE #423: "Actions" is not highlighted and expanded in menu when on page of action
Example: https://docs.fastlane.tools/actions/gym/

image

"Actions" in navigation is not expanded, "Building" is not highlighted.

janpio:
Currently "Actions" is not highlighted when you are on an action docs page:
https://docs.fastlane.tools/actions/sigh/

This PR changes that behavior so "Actions" is always highlighted when you are on one of these pages.

I also added some comments to the file, so it will hopefully be easier to understand what is going on in there next time someone has to do so.

closes #423

### DIFF
--- a/theme/toc.html
+++ b/theme/toc.html
@@ -9,8 +9,12 @@
   {% endif %}
 {% else %}
   {% if not nav_item.title.startswith("_") %}
-    <li class="toctree-l1 {% if nav_item.active%}current{%endif%}">
-      <a class="{% if nav_item.active%}current{%endif%}" href="{{ nav_item.url }}">{{ nav_item.title }}</a>
+    <li class="toctree-l1 {% if nav_item.active %}current{% endif %}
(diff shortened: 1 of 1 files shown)


theme/toc.html
```diff
@@ -1,24 +1,41 @@
+{# loop through all navigation items #}
+{# if the current item has children #}
 {% if nav_item.children %}
+  {# if the current item title does not start with _ #}
```
KrauseFx: Are those comments, the lines that start with `#`? Are they supposed to be in this PR?

janpio: Yes they are, and yes - I had a really hard time to understand the code in the beginning. Documented it for myself, and thought it might be nice to have that in general. 

Should I strip them?

janpio:
This does not work as expected:
https://deploy-preview-658--fastlane-docs-preview.netlify.com/actions/scan/
vs.
https://docs.fastlane.tools/actions/scan/

PR #663, opened 2018-08-03
Fix two typos in sync_code_signing.md decryption

provTheodoreNewell:
The example commands to decrypt the cert and generate an encrypted p12 are now self-consistent.

I'm not sure if this documentation is autogenerated from source elsewhere.


janpio:
It is, although not from where the file tells you ;) Systems are complicated...

The source of this file is in https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/sync_code_signing.md Would you mind creating a PR with the same changes over there please? Thanks.

provTheodoreNewell:
Thanks! I'll close this and make a new PR 👍

janpio:
(Mention me to get a super quick review and merge)

PR #673, opened 2018-08-10
Update "Collect your Google credentials", improve nested list, improve includes

janpio:
- https://github.com/fastlane/fastlane/issues/13082 triggered a review of the Google credentials creation process, I made some tiny improvements.
- The double list I created to make clearer that some parts happen in another window was not that pretty, so I fixed the CSS to not add a margin-bottom if we are in a nested list.
- So this doesn't have to be changed in 3 places, I extracted the list into an include.
   - And moved the includes to their own folder.


docs/actions/upload_to_play_store.md
```diff
@@ -46,20 +46,7 @@ _supply_ uploads app metadata, screenshots, binaries, and app bundles to Google
...
-1. Click **Select a role** and choose **Service Accounts > Service Account User**
-1. Check the **Furnish a new private key** checkbox
-1. Select **JSON** as the Key type
-1. Click **Create** to close the dialog
-1. Make a note of the file name of the JSON file downloaded to your computer
-1. Back on the Google Play developer console, click **Done** to close the dialog
-1. Click on **Grant Access** for the newly added service account
-1. Choose **Release Manager** from the **Role** dropdown
-1. Click **Add user** to close the dialog
+{!docs/includes/google-credentials.md!}
```
KrauseFx: I love that we can do that with `mkdocs` 👍

janpio: Absolutely, so useful.

PR #687, opened 2018-08-21
[WIP] [CI] Circle 2.0

janpio:
This PR creates the new, necessary Circle CI 2.0 configuration file, based on an auto generated conversion using https://circleci.com/docs/2.0/config-translation/ (https://circleci.com/api/v1.1/project/github/fastlane/docs/config-translation)

This (semi) works, but it is definitely not perfect:

## Problems

1. The auto generated config.yml seems to install everything just fine, but with Circle 1.0 we got this on `pipenv install`:

    > Courtesy Notice: Pipenv found itself running within a virtual environment, so it will automatically use that environment, instead of creating its own for any project.

    Now we don't, but this:

    > To activate this project's virtualenv, run `pipenv shell`.
    > Alternatively, run a command inside the virtualenv with `pipenv run`.

   This currently causes `bundle exec fastlane test` to fail with this:

    `[10:32:18]: No such file or directory - mkdocs`

    which makes sense, as we installed `mkdocs` in the virtual environment.

    I worked around this by replacing `mkdocs` with `pipenv run mkdocs` in the `Fastfile`. No ideas about the side effects.

1. Deployment is not covered by the conversion, 

   ```
    # Deployment
    # Your existing circle.yml file contains deployment steps.
    # The config translation tool does not support translating deployment steps
    # since deployment in CircleCI 2.0 are better handled through workflows.
    # See the documentation for more information https://circleci.com/docs/2.0/workflows/
   ```

    But we have this: https://github.com/fastlane/docs/blob/b81febfb7e9b981b95418b349e2c4a60dca4d1bd/circle.yml#L15-L19

   1. A possible solution might be to adapt this example to just run our shell script:

      ```
      - deploy:
          command: |
            if [ "${CIRCLE_BRANCH}" == "master" ]; then
              ansible-playbook site.yml
            fi
      ```
      https://circleci.com/docs/2.0/configuration-reference/#example-3

1. How can I test the `deployment` step on a branch/PR?

1. Our `deploy.sh` also uses `mkdocs` (problem see above) and creates a `circle.yml` file on the `gh-pages` branch so Circle can also run a pseudo build on the commits. What do we do about those?

1. We need both ruby and python, but all the docs assume I need one or the other. That is why this configuration takes so many more lines to achieve the same thing it seems - it uses a general purpose image and then installs and configures the things we need (and the code that was generated to do that was also broken - but I think I fixed it)


.circleci/config.yml
```diff
@@ -0,0 +1,119 @@
...
+    
+    # Prepare for artifact and test results  collection equivalent to how it was done on 1.0.
+    # In many cases you can simplify this from what is generated here.
+    # 'See docs on artifact collection here https://circleci.com/docs/2.0/artifacts/'
+    - run: mkdir -p $CIRCLE_ARTIFACTS $CIRCLE_TEST_REPORTS
+    
+    # This is based on your 1.0 configuration file or project settings
+    - run:
+        working_directory: ~/fastlane/docs
+        command: pyenv install --skip-existing 3.6.0 && pyenv global 3.6.0 && pyenv rehash && pip install virtualenv && pip install nose && pip install pep8 && pyenv rehash
```
DanToml: `- run: apt-get update && apt-get install python3.6 && pip install pyenv pipenv`

then you can also completely rm the rvm step below

.circleci/config.yml
```diff
@@ -0,0 +1,119 @@
...
+        keys:
+        # This branch if available
+        - v1-dep-{{ .Branch }}-
+        # Default branch if not
+        - v1-dep-master-
+        # Any branch if there are none on the default branch - this should be unnecessary if you have your default branch configured correctly
+        - v1-dep-
+
+    # This is based on your 1.0 configuration file or project settings
+    - run: pip install pipenv
```
DanToml: can rm with the above change

.circleci/config.yml
```diff
@@ -0,0 +1,119 @@
...
+    - run: pipenv install
+    
+    # The following line was run implicitly in your 1.0 builds based on what CircleCI inferred about the structure of your project. In 2.0 you need to be explicit about which commands should be run. In some cases you can discard inferred commands if they are not relevant to your project.
+    - run: echo -e "export RAILS_ENV=test\nexport RACK_ENV=test" >> $BASH_ENV
```
DanToml: can rm this

.circleci/config.yml
```diff
@@ -0,0 +1,119 @@
...
+    - run: 'bundle check --path=vendor/bundle || bundle install --path=vendor/bundle
+        --jobs=4 --retry=3 '
+
+    # Save dependency cache
+    - save_cache:
+        key: v1-dep-{{ .Branch }}-{{ epoch }}
```
DanToml: `v1-dep-{{ checksum Gemfile.lock }}` would be  a better cache key here and in the `restore_cache` step

.circleci/config.yml
```diff
@@ -0,0 +1,119 @@
...
+        - ~/.ivy2
+        - ~/.bundle
+        - ~/.go_workspace
+        - ~/.gradle
+        - ~/.cache/bower
+
+    # Test
+    #   This would typically be a build job when using workflows, possibly combined with build
+    # This is based on your 1.0 configuration file or project settings
+    - run: bundle update
```
DanToml: We shouldn't `bundle update` in CI

.circleci/config.yml
```diff
@@ -0,0 +1,119 @@
...
+    - run: bundle exec fastlane test
+    - run: bundle exec danger || echo "danger failed, moving on"
+    
+    # Deployment
```
DanToml: can rm this deployment and teardown commentary

.circleci/config.yml
janpio: both lines or just the one with the comment?

.circleci/config.yml
janpio: both lines or just the one with the comment?

.circleci/config.yml
janpio: Nope:

```
#!/bin/bash -eo pipefail
sudo apt-get update && apt-get install python3.6 && pip install pyenv pipenv

Ign:1 http://deb.debian.org/debian stretch InRelease

Get:2 http://deb.debian.org/debian stretch-updates InRelease [91.0 kB]

Hit:3 http://deb.debian.org/debian stretch Release

Get:4 http://security.debian.org/debian-security stretch/updates InRelease [94.3 kB]

Fetched 185 kB in 0s (549 kB/s)

Reading package lists... Done

E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
Exited with code 100
```
https://circleci.com/gh/fastlane/docs/1762?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

.circleci/config.yml
```diff
@@ -69,6 +69,7 @@ jobs:
     # This is based on your 1.0 configuration file or project settings
     - run: pip install pipenv
     - run: pipenv install
+    - run: pipenv shell
```
revolter: This should come before `pipenv install`. It still works as it is because `install` automatically creates the virtualenv, but it makes more sense to create and activate it ourselves (`pipenv shell`) if we already need this.

.circleci/config.yml
```diff
@@ -0,0 +1,99 @@
...
+    
+    # The following `checkout` command checks out your code to your working directory. In 1.0 we did this implicitly. In 2.0 you can choose where in the course of a job your code should be checked out.
+    - checkout
+    
+    # Prepare for artifact and test results  collection equivalent to how it was done on 1.0.
+    # In many cases you can simplify this from what is generated here.
+    # 'See docs on artifact collection here https://circleci.com/docs/2.0/artifacts/'
+    - run: mkdir -p $CIRCLE_ARTIFACTS $CIRCLE_TEST_REPORTS
+    
+    - run: sudo apt-get update && apt-get install python3.6 && pip install pyenv pipenv
```
revolter: `pyenv` is not needed.

revolter: @DanToml, Couldn't this be on separate lines? As it looks like it defeats the purpose of the ability to specify multiple `run`s.

janpio: (Currently this line crashes the build anyway - see my comment above)

janpio:
We are on a deadline now:

> Cloud-hosted CircleCI 1.0 service alert:
> 
> CircleCI 1.0 has been [EOL since September 2018](https://circleci.com/blog/sunsetting-1-0/). In order to continue to increase the efficiency of the CircleCI 2.0 platform, we will be shutting down the CircleCI 1.0 cloud-hosted fleet on March 15, 2019.
> 
> Starting March 15, cloud-hosted CircleCI 1.0 jobs will be viewable but will cease to run. For more information and a guide to migrating to CircleCI 2.0, please visit our [migration resource center](url).

The migration resource link in the email was broken :/

KrauseFx:
Ugh, probably easier to switch to Travis or something instead of getting Circle 2 to work 😂

janpio:
```
This is a follow-up reminding you that one or more of your projects uses CircleCI 1.0 configuration and will no longer build after March 15, 2019.

In order to address this issue and keep your project(s) building, please see:

    Instructions on seeing which projects still use CircleCI 1.0 config
    A guide to migrating from 1.0 to 2.0
    CircleCI 1.0 EOL policy

We know we have sent you a number of notifications in recent weeks regarding 1.0 EOL. At the risk of over-informing you, our goal is to make sure that zero teams or projects are caught unawares when they can no longer build with CircleCI 1.0 configuration on March 15, 2019.

We are here to help in making this switch, please visit the migration center.
```

PR #688, opened 2018-08-21
Run rubocop tests as part of CI

KrauseFx:


janpio:
Oh yeah, those new `rubocop` versions are difficult. And we both missed it @joshdholtz :/

PR #690, opened 2018-08-21
Add mdx_truly_sane_lists to Pipfile

KrauseFx:
I couldn't run the command to update the lock file due to broken python env on current machine, feel free to submit a separate PR for that

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -8,6 +8,7 @@ markdown-include = "==0.5.1"
 mdx-truly-sane-lists = "==1.0"
 mkdocs = "==0.15.3"
 pymdown-extensions = "==1.1"
+mdx_truly_sane_lists = "==1.1.1"
 
 [dev-packages]
 


Pipfile
```diff
@@ -8,6 +8,7 @@ markdown-include = "==0.5.1"
 mdx-truly-sane-lists = "==1.0"
 mkdocs = "==0.15.3"
 pymdown-extensions = "==1.1"
+mdx_truly_sane_lists = "==1.1.1"
```
revolter: It's so weird that I installed `mdx-truly-sane-lists` before (3 lines above). I don't know what the correct one is.

revolter: Did you add this by hand? Because running `pipenv install mdx_truly_sane_lists` reports: `Requirement already satisfied: mdx_truly_sane_lists`. What was the error you received?

KrauseFx: Oh yah I added this by hand, nice catch, will close PR 👍

KrauseFx:
My bad, thanks for catching the error

revolter:
Let me know if you find any issues with anything related to dependencies.

Issue #691, opened 2018-08-21
External links should always open in new tab


KrauseFx:
e.g. visit `/` landing page, and click on contributor face. This should not open in the same window, but in a new one. This should be automatic based on the host matching the current host. I use the following code successfully for krausefx.com

https://github.com/KrauseFx/krausefx.com/blob/5ffcbb50d944d5246781a078646ef615127957f1/_includes/footer.html#L23-L30

PR #703, opened 2018-08-22
Open external links in new tabs

KrauseFx:
Fixes https://github.com/fastlane/docs/issues/691


janpio:
Hm, what defines if a link gets `target=_blank?

First link on https://deploy-preview-703--fastlane-docs-preview.netlify.com/actions/#fastlane-actions (available actions)  opens in a new tab now for example.

PR #706, opened 2018-08-22
Remove superfluous default MkDocs CLI argument

revolter:


revolter:
Just because its the default and is a bit confusing. I thought it doesn't clean if I don't pass it locally.

```
  -c, --clean / --dirty           Remove old files from the site_dir before
                                  building (the default).
```

KrauseFx:
Perfect 🔢

PR #718, opened 2018-09-18
Make more code non-bold

KrauseFx:
screen shot 2018-09-18 at 11 45 50 am


KrauseFx:
Oh I actually haven't tried this, and it doesn't seem to work

PR #735, opened 2018-10-16
Make the plugin score look more beautiful

KrauseFx:
No idea why it was so ugly before 😂

screenshot 2018-10-16 03 42 32

---

### to

---

screenshot 2018-10-16 03 42 15


getaaron:
Sooo much better 🌈♥️🤤

PR #738, opened 2018-11-05
[match] update manual clone

joshdholtz:
Fixed manual clone docs that were using `GitHelper` which was deleted in https://github.com/fastlane/fastlane/pull/13405


getaaron:
🙏

PR #744, opened 2018-11-20
Add information about number of commits as build number

KrauseFx:


srsjaladi:
can you help me on teams(Microsoft) plugin with fastlane?

PR #759, opened 2018-12-11
Deliver submit build

joshdholtz:

### DIFF
--- a/docs/actions/deliver.md
+++ b/docs/actions/deliver.md
@@ -101,12 +101,6 @@ To get a list of available options run
 fastlane action deliver
 ```
 
-Select a previously uploaded build and submit it for review.
-
-```no-highlight
-fastlane deliver submit_build --build_number 830
-```
-
 ### Use in a `Fastfile`
 
 ```ruby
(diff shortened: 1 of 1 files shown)


docs/actions/deliver.md
```diff
@@ -406,6 +400,33 @@ The english name of the secondary first sub category you want to set
 The english name of the secondary second sub category you want to set
 </details>
 
+# Submit Build
+
+Select a previously uploaded build and submit it for review.
```
KrauseFx: To make it easier for search, maybe add something like `Promote an existing build to production` or similar as well as what we already have 👍

docs/actions/deliver.md
```diff
@@ -406,6 +400,33 @@ The english name of the secondary first sub category you want to set
...
+
+    force: true, # Skip HTMl report verification
+    skip_metadata: true,
+    skip_screenshots: true,
+    skip_binary_upload: true
+  )
+end
+```
+
+Omit `build_number` to let _fastlane_ automatically select the current edit version from App Store Connect.
```
KrauseFx: `edit version` is probably something people are not aware of, can we rephrase?

PR #772, opened 2019-02-03
Android Setup: Add information about Java and Gradle installation

asterixorobelix:
As discussed in my recent [issue](https://github.com/fastlane/fastlane/issues/14135)

### DIFF
--- a/docs/getting-started/android/setup.md
+++ b/docs/getting-started/android/setup.md
@@ -12,8 +12,10 @@ Android Studio maintains its own instances of Gradle and Java, which you do not 
 For macOS, the easiest way to install Java 8 is with [Homebrew](https://brew.sh/).
 
 Run the following commands in the terminal:
-`brew tap caskroom/versions`
-`brew cask install java8`
+```
+brew tap caskroom/versions
+brew cask install java8
+```
 
 The terminal command: `javac -version` should now return something similar to: `javac 1.8.0_202-ea`
 
--- a/docs/getting-started/android/setup.md
+++ b/docs/getting-started/android/setup.md
@@ -9,7 +9,7 @@
 Android Studio maintains its own instances of Gradle and Java, which you do not have access to in the command line. So even though you can build and run apps in Android Studio, you need to install those tools for fastlane separately. Additionally, you need to install Java _8_ and not a later version of Java, which do not play well with Gradle.
 
 ### Install Java 8
-For MacOS, the easiest way to install Java 8 is with [Homebrew](https://brew.sh/).
+For macOS, the easiest way to install Java 8 is with [Homebrew](https://brew.sh/).
 
 Run the following commands in the terminal:
 `brew tap caskroom/versions`
@@ -17,7 +17,7 @@ Run the following commands in the terminal:
 
 The terminal command: `javac -version` should now return something similar to: `javac 1.8.0_202-ea`
 
-If this is not the case, your JAVA_HOME environment variable likely refers to a more recent version of Java. The easiest way to fix this is to uninstall all java instances via Homebrew with `brew cask uninstall java` and then install java 8 with `brew cask install java8`
+If this is not the case, your JAVA_HOME environment variable likely refers to a more recent version of Java. The easiest way to fix this is to uninstall all Java instances which were installed via Homebrew with `brew cask uninstall java` and then install Java 8 with `brew cask install java8`
 
 ### Install Gradle
 
--- a/docs/getting-started/android/setup.md
+++ b/docs/getting-started/android/setup.md
@@ -6,7 +6,7 @@
 
 ## Setting up _fastlane_
 
-Android Studio maintains its own instances of Gradle and Java, which you do not have access to in the command line, so even though you can build and run apps in Android Studio, you need to install those tools for fastlane. Additionally, you need to install Java 8 and not later versions of Java, which do not play well with Gradle.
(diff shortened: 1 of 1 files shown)


asterixorobelix:
@janpio

docs/getting-started/android/setup.md
```diff
@@ -6,6 +6,25 @@
...
+`brew tap caskroom/versions`
+`brew cask install java8`
+
+The terminal command: `javac -version` should now return `javac 1.8.0_202-ea`
+
+If this is not the case, your JAVA_HOME environment variable likely refers to a more recent version of Java. The easiest way to fix this is to uninstall all java instances via Homebrew with `brew cask uninstall java` and then install java 8 with `brew cask install java8`
+
+### Install Gradle
+
+on MacOs you can install Gradle via homebrew with the command `brew install gradle`
```
janpio:
```suggestion
On macOS you can install Gradle via homebrew with the command `brew install gradle`.
```

asterixorobelix:
I have signed the CLA

docs/getting-started/android/setup.md
```diff
@@ -6,6 +6,25 @@
...
+### Install Java 8
+For MacOS, the easiest way to install Java 8 is with [Homebrew](https://brew.sh/).
+
+Run the following commands in the terminal:
+`brew tap caskroom/versions`
+`brew cask install java8`
+
+The terminal command: `javac -version` should now return `javac 1.8.0_202-ea`
+
+If this is not the case, your JAVA_HOME environment variable likely refers to a more recent version of Java. The easiest way to fix this is to uninstall all java instances via Homebrew with `brew cask uninstall java` and then install java 8 with `brew cask install java8`
```
janpio:
```suggestion
If this is not the case, your `JAVA_HOME` environment variable likely refers to a more recent version of Java. The easiest way to fix this is to uninstall all Java instances via Homebrew with `brew cask uninstall java` and then install Java 8 with `brew cask install java8`.
```

Can you actually uninstall Java versions not installed via Homebrew with it?

docs/getting-started/android/setup.md
```diff
@@ -6,6 +6,25 @@
...
+Android Studio maintains its own instances of Gradle and Java, which you do not have access to in the command line, so even though you can build and run apps in Android Studio, you need to install those tools for fastlane. Additionally, you need to install Java 8 and not later versions of Java, which do not play well with Gradle.
+
+### Install Java 8
+For MacOS, the easiest way to install Java 8 is with [Homebrew](https://brew.sh/).
+
+Run the following commands in the terminal:
+`brew tap caskroom/versions`
+`brew cask install java8`
+
+The terminal command: `javac -version` should now return `javac 1.8.0_202-ea`
```
janpio: Exactly this version or similar?

docs/getting-started/android/setup.md
```diff
@@ -6,6 +6,25 @@
 
 ## Setting up _fastlane_
 
+Android Studio maintains its own instances of Gradle and Java, which you do not have access to in the command line, so even though you can build and run apps in Android Studio, you need to install those tools for fastlane. Additionally, you need to install Java 8 and not later versions of Java, which do not play well with Gradle.
```
janpio:
```suggestion
Android Studio maintains its own instances of Gradle and Java, which you do not have access to in the command line. So even though you can build and run apps in Android Studio, you need to install those tools for fastlane separately. Additionally, you need to install Java _8_ and not a later version of Java, which do not play well with Gradle.
```

docs/getting-started/android/setup.md
asterixorobelix: Something similar. I'll update

docs/getting-started/android/setup.md
asterixorobelix: I installed the latest Java via Homebrew before I discovered that it was the incorrect version and then needed to uninstall it. I imagine you can only uninstall the Homebrew Java versions.

asterixorobelix:
@janpio It seems we need your CLA consent too

asterixorobelix:
@janpio You are now an author on this because I merged your recommendations in. Need your CLA

docs/getting-started/android/setup.md
```diff
@@ -6,6 +6,27 @@
...
+
+Run the following commands in the terminal:
+```
+brew tap caskroom/versions
+brew cask install java8
+```
+
+The terminal command: `javac -version` should now return something similar to: `javac 1.8.0_202-ea`
+
+If this is not the case, your JAVA_HOME environment variable likely refers to a more recent version of Java. The easiest way to fix this is to uninstall all Java instances which were installed via Homebrew with `brew cask uninstall java` and then install Java 8 with `brew cask install java8`
```
janpio:
```suggestion
If this is not the case, your `JAVA_HOME` environment variable likely refers to a more recent version of Java. The easiest way to fix this is to uninstall all Java instances which were installed via Homebrew with `brew cask uninstall java` and then install Java 8 with `brew cask install java8`
```

janpio:
My consent is implied, @googlebot just can't handle it properly. As I don't work for Google, I can also not just set the label to `cla: yes` but I can later just merge this and ignore what the label says.

janpio:
Are Java and Gradle needed for _all_ the users of fastlane for Android? If not, it might be better to actually move this further down somewhere and mention what action actually needs them. No need to make fastlane installation or usage harder than it has to be, right?

rogerluan:
@asterixorobelix ping 🤗

rogerluan:
Friendly ping @asterixorobelix 🙈

PR #773, opened 2019-02-05
Added Bitrise / Fastlane doc

RikH81:
Added guide to Fastlane integration with Bitrise CI

PR #774, opened 2019-02-05
Added Bitrise to the list

RikH81:
Pretty much what it says on the tin. Should only happen after the actual guide's committed though :)


RikH81:
Just now signed the CLA for Bitrise <-

janpio:
Hey @RikH81, seems you are actually missing the CLA here that is required that we can merge this. This also points to a problem for #773 where I assumed the CLA problem was because you had accepted one of my change suggestions, but seems the problem is a deeper one.

Could you please fix that?

rogerluan:
Closing, as Bitrise is now already present in _fastlane_ docs: https://docs.fastlane.tools/best-practices/continuous-integration/bitrise/

PR #778, opened 2019-02-13
Apple forcing 2FA disables solution proposed on Best Practices » Continuous Integration

marcosgriselli:
2famail

The best practices on continuous integration docs recommend creating a new Apple ID without 2FA enabled when submitting builds from CI. This approach will no longer work on February 27th, 2019 since Apple will enforce 2FA on developers Apple ID. The docs should be updated to remove this recommendation once 2FA is obligatory.

I'd update the docs but I'm not sure what's the recommended setup when

### DIFF
--- a/docs/best-practices/continuous-integration.md
+++ b/docs/best-practices/continuous-integration.md
@@ -20,16 +20,7 @@ Multiple CI products and services offer integrations with fastlane:
 
 ### Separate Apple ID for CI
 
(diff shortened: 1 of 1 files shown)


janpio:
Thanks for this PR @marcosgriselli, I created #779 with some more elaborate changes. Could you please look over those to see if they make sense to you? Thanks.

marcosgriselli:
@janpio that looks a lot better. I'll close this one and add comments there.

PR #782, opened 2019-02-20
[codesigning/xcode-project] use `xcargs` instead of `export_xcargs`

PatrickDotStar:
`export_xcargs` does not work for Xcode 10. `xcargs` should be used instead


PatrickDotStar:
I signed it!

PR #787, opened 2019-02-21
add that you need to have done an ad-hoc apk upload before supply init

asterixorobelix:

### DIFF
--- a/docs/getting-started/android/setup.md
+++ b/docs/getting-started/android/setup.md
@@ -33,6 +33,8 @@ The most interesting file is `fastlane/Fastfile`, which contains all the informa
 
 _supply_ is a _fastlane_ tool that uploads app metadata, screenshots and binaries to Google Play. You can also select tracks for builds and promote builds to production!
 
+For _supply_ to be able to initialize, you need to have successfully uploaded an apk to you Google Play Console app at least once.
+
 Setting it up requires downloading a credentials file from your Google Developers Service Account.
 
 ### Collect your Google credentials


janpio:
Very good point.

I thought we already had mentioned that somewhere though... maybe only on the `supply` page itself and not on the setup?

asterixorobelix:
@janpio this is documentation, no need to be DRY. In my opinion reiteration is good in documentation. this point was missed by me when I was reading the docs

docs/getting-started/android/setup.md
```diff
@@ -33,6 +33,8 @@ The most interesting file is `fastlane/Fastfile`, which contains all the informa
 
 _supply_ is a _fastlane_ tool that uploads app metadata, screenshots and binaries to Google Play. You can also select tracks for builds and promote builds to production!
 
+For _supply_ to be able to initialize, you need to have successfully uploaded an apk to you Google Play Console app at least once.
```
janpio:
```suggestion
For _supply_ to be able to initialize, you need to have successfully uploaded an APK to your app in the Google Play Console at least once.
```

janpio:
That wasn't a criticism, I was just trying to get the full picture.  
If we have similar wording somewhere else, we should make sure it is consistent.

docs/getting-started/android/setup.md
rogerluan:
```suggestion
For _supply_ to be able to initialize, you need to have successfully uploaded an APK to your app in the Google Play Console at least once.
```

rogerluan (changes requested):

rogerluan:
Closing in favor of https://github.com/fastlane/docs/pull/1030

PR #797, opened 2019-03-08
Make app specific password requirements more clear

mpdifran:
- Add more clarity about which cases allow the use of an application specific password, 
specifically in regards to TestFlight commands.
- Fix spelling mistake for command upload_to_testflight

This is a result of the conversation here https://github.com/fastlane/fastlane/issues/14365

### DIFF
--- a/docs/best-practices/continuous-integration.md
+++ b/docs/best-practices/continuous-integration.md
@@ -60,7 +60,7 @@ If you want to upload builds to App Store Connect (actions `upload_to_app_store`
 
 This will supply the application specific password to iTMSTransporter, the tool used by those actions to perform the upload.
 
-It should be noted that the application specific password will only work for uploading the binary; it won't work for setting release notes or distributing to testers, etc.
+Note: The application specific password will _not_ work if your action usage does anything else than uploading the binary, e.g. updating any metadata like setting release notes or distributing to testers, etc.
 
 ##### `spaceauth`
 
--- a/docs/best-practices/continuous-integration.md
+++ b/docs/best-practices/continuous-integration.md
@@ -52,13 +52,15 @@ When you can not enter the security code manually, as on a Continuous Integratio
 
 ##### Application specific passwords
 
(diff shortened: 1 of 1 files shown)


janpio:
Thanks for these changes, the typo fix is definitely valid.

The other part is problematic though - it is currently in the second set of parenthesises which only makes it apply to `pilot`, which is not what we want to say here.

I am confused right now when we can actually use the app specific password... Does it only work if we _just_ upload a binary? Or does it also work when uploading a binary _and_ update metadata, but which then uses the other authentication mechanism(s)? So confusing :/

mpdifran:
Ah you're right, it should probably apply outside the parentheses. It's my understanding that the ASP only works for uploading a binary for all these commands, and won't work for metadata updates, etc.

janpio:
yes, as soon as there is also metadata stuff it won't work with _only_ the ASP any more. But you might still use it for the binary-upload-part I think. "Just" the metadata part will require the normal auth mess I think.

mpdifran:
Anything else you'd like me to change?

docs/best-practices/continuous-integration.md
```diff
@@ -52,14 +52,16 @@ When you can not enter the security code manually, as on a Continuous Integratio
...
-If you want to upload builds to App Store Connect (actions `upload_to_app_store` and `deliver`) or TestFlight (actions `upload_to testflight`, `pilot` or `testflight`) from your CI machine, you need to generate an _application specific password_:
+If you want to upload builds to App Store Connect (actions `upload_to_app_store` and `deliver`) or TestFlight (actions `upload_to_testflight`, `pilot` or `testflight`) from your CI machine, you need to generate an _application specific password_:
 
 1. Visit [appleid.apple.com/account/manage](https://appleid.apple.com/account/manage)
 1. Generate a new application specific password
 1. Provide the application specific password using the environment variable `FASTLANE_APPLE_APPLICATION_SPECIFIC_PASSWORD`
 
 This will supply the application specific password to iTMSTransporter, the tool used by those actions to perform the upload.
 
+It should be noted that the application specific password will only work for uploading the binary; it won't work for setting release notes or distributing to testers, etc.
```
janpio:
```suggestion
Note: The application specific password will _not_ work if your action usage does anything else than uploading the binary, e.g. updating any metadata like setting release notes or distributing to testers, etc.
```

janpio:
Ah, didn't see you change. Suggested a rewrite, see if you like that one.

PR #802, opened 2019-03-20
Migrate to circle 2.0

joshdholtz:
🤞


getaaron:
I don’t think just renaming will work, I think the syntax is different

joshdholtz:
@getaaron lol, I know 😉 I just wanted to see if it would run and what the error was

.circleci/config.yml
```diff
@@ -0,0 +1,39 @@
...
+  "Test":
+    docker: # Replace this with a fastlane owned docker image at some point
+      - image: joshdholtz/fastlane-docs:0.0.1
+    environment:
+      LC_ALL: C.UTF-8
+      LANG: C.UTF-8
+    working_directory: ~/code
+    steps:
+      - checkout
+      - run: pip3 install pipenv
```
janpio: `pip` (circle.yml) vs. `pip3` - What the difference?

.circleci/config.yml
```diff
@@ -0,0 +1,39 @@
...
+      - run: pipenv install
+      - run: bundle update
+      - run: pipenv run bundle exec fastlane test
```
janpio: In the old one this command was not run via `pipenv run` - what's the difference?

.circleci/config.yml
```diff
@@ -0,0 +1,39 @@
+version: 2.0
+
+jobs:
+  "Test":
+    docker: # Replace this with a fastlane owned docker image at some point
+      - image: joshdholtz/fastlane-docs:0.0.1
+    environment:
```
janpio: Could this maybe be moved to the docker image so it doesn't have to be repeated here?

.circleci/images/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ruby:2.3.4
+
+ARG BUILDDIR="/tmp/build"
+ARG PYTHON_VER="3.6.8"
+WORKDIR ${BUILDDIR}
+
+RUN apt-get update -qq && \
+apt-get upgrade -y  > /dev/null 2>&1 && \
```
janpio:
```suggestion
apt-get upgrade -y > /dev/null 2>&1 && \
```

janpio:
Left some comments.

This has to be taken care of as well I guess: https://github.com/fastlane/docs/blob/6091a1b57513fa177626d62b43c79cc562477ff5/scripts/ci/deploy.sh#L45-L48 (Although I have no idea why this is needed in the first place really.)

.circleci/config.yml
getaaron: `pip3` is for python 3. Guessing the newer 2.0 images have newer python? 🐍

.circleci/config.yml
joshdholtz: We need access to `mkdocs` and `mkdocs` is only available inside of a `pipenv shell`. Since are we in a CircleCI shell already we can't enter a `pipenv shell` so we have to do it on a command by command basis. Without `pipenv run` it errors with `mkdocs not found`.

.circleci/config.yml
joshdholtz: Can do can do!

.circleci/config.yml
joshdholtz: I made this newer image have python 3 so `pip3` is what we need here 😊

joshdholtz:
@janpio Made your requested changes! Going to merge this and see how the deploy goes 🙃

.circleci/config.yml
janpio: Ah, I remember those from when I tried to migrate. Pain in the ass.

janpio:
should have squash merged this 🙈

PR #806, opened 2019-03-24
Added app_review_attachment_file documentation.

krish722:
I am adding documentation for the following PR

https://github.com/fastlane/fastlane/pull/14412

and the feature developed after this request.
https://github.com/fastlane/fastlane/issues/14285

If more details needed, please let me know.


janpio:
This documentation is actually generated from https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/upload_to_app_store.md.erb, you should add those changes there (either in a new PR or include it in your existing PR).

janpio:
If it's specifically for this change, you will have to check the file in the main repo. But in general we are happy to accept all PRs with fixups not matter how small :)

PR #809, opened 2019-03-27
Using our new legit dockerfile

joshdholtz:
Replacing with our legit dockerfile


.circleci/config.yml
```diff
@@ -2,12 +2,12 @@ version: 2.0
...
   "Test":
-    docker: # Replace this with a fastlane owned docker image at some point
-      - image: joshdholtz/fastlane-docs:0.0.1
+    docker:
+      - image: fastlanetools/ci:0.1.0
     working_directory: ~/code
     steps:
       - checkout
-      - run: pip3 install pipenv
+      - run: sudo pip3 install pipenv
```
janpio: Why the new `sudo`?

PR #820, opened 2019-04-23
[frameit] Add missing documentation on title_min_height option

bvirlet:


janpio:
Thanks, and sorry, but similar to https://github.com/fastlane/fastlane/pull/14644#issuecomment-485809434 this has to be edited in the main fastlane repo.

PR #827, opened 2019-05-08
Remove Xcode 9 limitation

schayes04:
Running multiple simulators at once is supported on Xcode 10 as well. I removed the limitation note since by now everyone should be at least on Xcode 9 if not 10


schayes04:
I signed it!

getaaron:
Thanks!

PR #834, opened 2019-05-22
Fix display of markdown list

adrm:
The list must be in a new paragraph to be formatted as list and not as part of the previous paragraph.

This is how it is rendered without this patch, as you can see live in the docs: https://docs.fastlane.tools/getting-started/android/release-deployment/

imagen


adrm:
CLA signed.

janpio:
Thanks!

PR #847, opened 2019-06-28
Adds a hint that "setup_travis" is not needed for Android

bbuhler:
It took me a while until I found out why my build was failing with exit code 127.

I interpreted the documentation like that: if I use fastlane with Travis CI I need to call the "setup_travis" action and was wondering why it's only showing an example for iOS.

To save someone the same trouble I added a note at the end of file.

### DIFF
--- a/docs/best-practices/continuous-integration/travis.md
+++ b/docs/best-practices/continuous-integration/travis.md
@@ -22,3 +22,5 @@ setup_travis
 ```
 
 which will setup the keychain to work well with _match_ and _gym_. This action will also enable the `readonly` mode for _match_, so your CI doesn't create new certificates or provisioning profiles.
+
+Note: This is only required for iOS and not for Android.


janpio:
> It took me a while until I found out why my build was failing with exit code 127.

Why did it? Caused by `setup_travis`?

bbuhler:
I assume so, yes.

https://www.travis-ci.org/bbuhler/Todo2Wish/builds/551959305#L1457

As you can see it is running "setup_travis" and fails afterwards.
After I removed the action it was working.

janpio:
Ah, Android/Ubuntu environment on Travis. We should probably check that the `setup_ci` methods better handle non macOS platforms.

mollyIV:
👋 @bbuhler, 

> I interpreted the documentation like that: if I use fastlane with Travis CI I need to call the "setup_travis" action and was wondering why it's only showing an example for iOS.

Do you think the issue might be that we do not encourage a user to check `setup_travis` action documentation in a [`Travis Integration`](https://docs.fastlane.tools/best-practices/continuous-integration/travis/) section to get more details, i.e. putting a link to `setup_travis` documentation might help? 🤔 Then it would be way easier to discover supported platforms - there is a `Supported platforms` table at the very top: https://docs.fastlane.tools/actions/setup_travis/. 

> Ah, Android/Ubuntu environment on Travis. We should probably check that the setup_ci methods better handle non macOS platforms.

@janpio I had a quick look at the code, and it seems that we only support `ios` and `mac` platforms for CIs. 

```
def self.is_supported?(platform)
        [:ios, :mac].include?(platform)
end
```
so not sure how we could improve that 🤔 Maybe we shall add a note with platforms restrictions to [Best practices: Continuous Integration documentation](https://docs.fastlane.tools/best-practices/continuous-integration/)?

janpio:
I am not sure how we want to handle something like that. We could also change the action to just work (and do nothing) for Android projects.

bbuhler:
> Do you think the issue might be that we do not encourage a user to check setup_travis action documentation in a Travis Integration section to get more details, i.e. putting a link to setup_travis documentation might help?

@mollyIV I think that would work as well, though the information is still spread over two documentations.

I can adapt my PR to add a link to setup_travis, if that is fine for you.

> We should probably check that the setup_ci methods better handle non macOS platforms.

@janpio I like that approach as well. Maybe we can print a message like the one you see when you run fastlane without Travis environment.

I can create an issue in https://github.com/fastlane/fastlane, if you like.

PR #853, opened 2019-07-12
Add note that FASTLANE_HIDE_TIMESTAMP is overruled by --verbose

janpio:
See https://github.com/fastlane/fastlane/blob/f32b007ff45e648b37b6c9c2037ac481f36b7780/fastlane_core/lib/fastlane_core/ui/implementations/shell.rb#L33-L39


janpio:
Thanks for the reviews.

PR #863, opened 2019-08-27
docs(react-native): Add new link to tutorial using detox for screenshotting

BatDroid:
A new link is added as a tutorial for having automatic screenshots in a react native project with Fastlane & Detox

### DIFF
--- a/docs/getting-started/cross-platform/react-native.md
+++ b/docs/getting-started/cross-platform/react-native.md
@@ -11,3 +11,4 @@ As there is no official getting started guide yet, here are some publications by
 - Team Focused Fastlane Tutorials
   - [Simple Android Releases in React Native](https://shift.infinite.red/simple-react-native-android-releases-319dc5e29605)
   - [Simple iOS Releases in React Native](https://shift.infinite.red/simple-react-native-ios-releases-4c28bb53a97b)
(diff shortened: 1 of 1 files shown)


BatDroid:
> @googlebot I signed it!

@googlebot I signed it!

janpio:
Interesting. `screengrab` didn't work because its automation doesn't work well with React Native? https://docs.fastlane.tools/actions/screengrab/#ui-tests

BatDroid:
> Interesting. `screengrab` didn't work because its automation doesn't work well with React Native? https://docs.fastlane.tools/actions/screengrab/#ui-tests

what i understood from the auto screenshot for both ios & android(screengrab) was that they needed native setup, linking, etc. so it wasn't very clear for a react native developer who want to have auto screenshot with minimum effort.

janpio:
That sounds very reasonable and is an avenue we should look into further from fastlane itself.

BatDroid:
> That sounds very reasonable and is an avenue we should look into further from fastlane itself.

so do you see this pr/article good enough to be merged until fastlane provide something for it?

joshdholtz (approved):
This looks good to me! Thanks for the contribution and sorry about the slow review 🤦‍♂️

PR #872, opened 2019-10-12
Add no_checkout option to carthage action

att55:
### related PR

https://github.com/fastlane/fastlane/pull/15482

### overview

Latest (ver: 0.33.0) carthage boot strap command has `--no-checkout` option.
But carthage action isn't implemented the option.
So, I made PR to add the option. I'll also update docs.


janpio:
The docs will updated automatically when your PR is merged and released - the complete documentation is generated based on the code of the action themselves. Thanks anyway!

PR #877, opened 2019-10-22
Plugins - add more clarity for using `bundle exec` after installing plugins

joshdholtz:
Helps fix https://github.com/fastlane/fastlane/issues/15506
Helps fix https://github.com/fastlane/fastlane/issues/15506[

Not clear to users will now have to use `bundle exec` after installing plugins 😊


lyndsey-ferguson:
Thanks for accepting my suggestion!

PR #879, opened 2019-11-05
Mention `fastlane update_plugins`

ricardopereira:
Demonstrate the available command to update plugins: `fastlane update_plugins`.


ricardopereira:
@googlebot I signed it!

janpio:
Thanks, very good catch.

Seems before we only mentioned this on https://docs.fastlane.tools/plugins/plugins-troubleshooting/#update-your-plugins

PR #896, opened 2020-01-08
Updated brew installation instructions

alexrepty:
According to [this note](https://github.com/fastlane/fastlane/issues/15496#issuecomment-565028879) by @joshdholtz, the old method of using `brew cask install fastlane` doesn't work anymore, as it installs an old version. The currently supported way is `brew install fastlane` (sans `cask`), which is not reflected in the docs just yet.

### DIFF
--- a/docs/includes/installing-fastlane.md
+++ b/docs/includes/installing-fastlane.md
@@ -11,5 +11,5 @@ Install _fastlane_ using
 sudo gem install fastlane -NV
 
 # Alternatively using Homebrew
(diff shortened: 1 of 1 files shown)


joshdholtz:
Ah, thanks for making this. Will look at this right away!!

alexrepty:
@googlebot I signed it!

joshdholtz:
Ah sorry! Closing this in favor of #895 which was made before his one 😔 I appreciate the contribution though!

PR #902, opened 2020-01-21
Update Xcode version to something more recent

BytesGuy:
Updated the Xcode version in this example to something more recent as Xcode 9 is pretty old by now 🙂


BytesGuy:
@googlebot I signed it!

PR #905, opened 2020-02-03
best practices ci: Update Xcode version

BytesGuy:
Xcode 9 is very old now and will likely be deprecated in the not too distant future, so updating the example here to 11.3


janpio:
Thanks, that looks useful.

PR #912, opened 2020-03-13
Update remaining brew install commands away from cask

clodargh:
Following up on https://github.com/fastlane/docs/pull/895 which missed a couple of other references to `brew cask install fastlane`.

The context of the [original issue](https://github.com/fastlane/fastlane/issues/15496#issuecomment-565028879) is that installing via `cask` leads to installation of an old version (2.28.3) of fastlane.

PS thanks for maintaining this tool!

### DIFF
--- a/docs/includes/setup-fastlane-header.md
+++ b/docs/includes/setup-fastlane-header.md
@@ -40,7 +40,7 @@ xcode-select --install
 sudo gem install fastlane -NV
 
 # Alternatively using Homebrew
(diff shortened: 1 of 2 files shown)


CloudyML:
^ That CLA looks like it's going to be a pain while I have the private email address setting enabled on GitHub, so please feel welcome to just roll this into another PR and close.

rogerluan:
These changes are already present in master 😊 closing! Thanks!

PR #932, opened 2020-06-30
add keychain troubleshoot explanation when `login.keychain` is not the default keychain

BenLorantfy:
I was running into an issue using GitHub actions where fastlane was using the system keychain instead of the login keychain.  It didn't have access to the system keychain so some commands would fail.

### DIFF
--- a/docs/codesigning/troubleshooting.md
+++ b/docs/codesigning/troubleshooting.md
@@ -49,7 +49,7 @@ If a certificate gets revoked, all connected provisioning profiles get invalidat
   <img src="/img/codesigning/KeychainPrivateKey.png" width=500 />
 </p>
 1. Make sure to have deleted all expired WWDR certificates, more information [here](https://stackoverflow.com/questions/32821189/xcode-7-error-missing-ios-distribution-signing-identity-for/35401483#35401483). There might be 2 expired WWDR certificates, one in the `login`, and one in the `system` keychain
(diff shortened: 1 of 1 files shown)


BenLorantfy:
@googlebot I signed it!

docs/codesigning/troubleshooting.md
```diff
@@ -49,6 +49,7 @@ If a certificate gets revoked, all connected provisioning profiles get invalidat
   <img src="/img/codesigning/KeychainPrivateKey.png" width=500 />
 </p>
 1. Make sure to have deleted all expired WWDR certificates, more information [here](https://stackoverflow.com/questions/32821189/xcode-7-error-missing-ios-distribution-signing-identity-for/35401483#35401483). There might be 2 expired WWDR certificates, one in the `login`, and one in the `system` keychain
+1. Run `security default-keychain` to view the default keychain that is used by some fastlane actions if a keychain path is not explicitly provided.  If it's the system keychain, make sure fastlane has access to write to it.  You may want to reconfigure your enviroment so the default keychain is the login keychain, which is less prone to permissions errors.  If you're using launchctl, see [this link to change the default keychain to the login keychain](https://serverfault.com/a/371252).
```
rogerluan, line 52:
```suggestion
1. Run `security default-keychain` to view the default keychain that is used by some _fastlane_ actions if a keychain path is not explicitly provided.  If it's the system keychain, make sure _fastlane_ has access to write to it.  You may want to reconfigure your enviroment so the default keychain is the login keychain, which is less prone to permissions errors.  If you're using `launchctl`, see [this link to change the default keychain to the login keychain](https://serverfault.com/a/371252).
```

rogerluan (approved):
Makes sense! Thanks for sharing your experience and a helpful link to a possible solution 🤗

PR #948, opened 2020-07-26
use `configuration` option instead of `scheme` in lanes.md sample code

chkpnt:
_Release_ and _Staging_ are most likely Xcode build configurations and not schemes.


chkpnt:
@googlebot I signed it!

rogerluan (approved):
Good point! Thanks for this @chkpnt 🤗

PR #951, opened 2020-08-04
Updating info regarding two-factor authentication

afa7789:
I've tried all the afternoon create an account without two factor and encountered with the apple support info regarding it that will be helpfull for future developers who are trying to continuous integrate as i've had.


afa7789:
@googlebot I signed it!

afa7789:
:)

rogerluan:
@afa7789 appreciate your contribution, but at this point I don't think this should be added to the documentation, especially because Apple has been blocking all non-2FA-compliant accounts from signing in since March 3rd 2021.

The documentation should be drastically updated for all the latest changes to 2FA, but that's should be done as a separate PR (which, by the way, would be very welcome! 🙏 )

Closing this PR. 👍

PR #965, opened 2020-08-31
Add note about requiring apple_id when using pilot with app specific password

ndonald2:
The pilot action requires an explicitly provided value for `apple_id` when using the application specific password auth strategy, otherwise it attempts to create an auth session with App Store Connect. Presumably to look up the value for `apple_id`.

This would have saved me hours of debugging trying to get the app specific password to work, and I would like to spare others from that going forward. The clue I needed was in this comment https://github.com/fastlane/fastlane/issues/17062#issuecomment-676612397.

### DIFF
--- a/docs/best-practices/continuous-integration.md
+++ b/docs/best-practices/continuous-integration.md
@@ -62,7 +62,7 @@ If you want to upload builds to App Store Connect (actions `upload_to_app_store`
 
 This will supply the application specific password to iTMSTransporter, the tool used by those actions to perform the upload.
 
-Note: The application specific password will _not_ work if your action usage does anything else than uploading the binary, e.g. updating any metadata like setting release notes or distributing to testers, etc.
+Note: The application specific password will _not_ work if your action usage does anything else than uploading the binary, e.g. updating any metadata like setting release notes or distributing to testers, etc. Additionally, when using `pilot`/`upload_to_testflight`, you _must explicitly provide a value for `apple_id`_, otherwise the action requires an auth session and will not use the application specific password.
 
 ##### `spaceauth`
 


ndonald2:
@googlebot I signed it!

rogerluan (commented):
I'm concerned about where this change was made. Not sure if Continuous Integration page should be where this information should live 🤔

ndonald2:
@rogerluan I'd be happy to move this to another document if you feel it fits better somewhere. I can't comment about "where this change was made" but as of 8/2020 when I opened this PR it was 100% the case that `apple_id` must be explicitly provided otherwise the App Specific Password will not work.

I'd also note that `apple_id` is confusingly named here - it's not a user account as you might expect, it's an identifier associated with the app itself. As per `pilot` docs:

```
apple_id: Apple ID property in the App Information section in App Store Connect
```

rogerluan:
When I read this PR the first time, I thought the CI page wasn't the best place to put this sort of information because it's very specific to `pilot`, and not strictly related to CI. For this reason I think it should live in `pilot`/`upload_to_testlight` documentation pages.

In fact, after searching for this info, turns out both `pilot` and `upload_to_testflight` docs explicitly mention this behavior:

http://docs.fastlane.tools/actions/upload_to_testflight/#use-an-application-specific-password-to-upload

http://docs.fastlane.tools/actions/pilot/#use-an-application-specific-password-to-upload

>_pilot_/`upload_to_testflight` can use an [Application Specific Password via the `FASTLANE_APPLE_APPLICATION_SPECIFIC_PASSWORD` environment variable](https://docs.fastlane.tools/best-practices/continuous-integration/#application-specific-passwords) to upload a binary if both the `skip_waiting_for_build_processing` and `apple_id` options are set. (If any of those are not set, it will use the normal Apple login process that might require 2FA authentication.)

I think we can close this PR since this info/behavior is already documented 😃 I appreciate your collaboration - let me know if you have further questions! 🤗

Ankit13Rid:
Circle CI still asking for Password in F2A. IDK how to resolve this now.

PR #967, opened 2020-09-01
Update brew cask install to brew install

ayozebarrera:
Brew cask install doesn't exists


ayozebarrera:
@googlebot I signed it!

ainame:
This seemed to be fixed already 10d96d0fe but thank you for taking the time 🙇

PR #975, opened 2020-09-14
Use brew instead of brew cask

lcd1232:


lcd1232:
@joshdholtz review plz

joshdholtz (approved):
Love it! Thanks for fixing 😊 Really appreciate the contribution ❤️

PR #992, opened 2020-10-14
Update app-store-connect-api.md

garrettmoon:
I think this is supposed to be key_content, not key: https://docs.fastlane.tools/actions/app_store_connect_api_key/

### DIFF
--- a/docs/app-store-connect-api.md
+++ b/docs/app-store-connect-api.md
@@ -88,7 +88,7 @@ Below is an example of the _fastlane_ API Key JSON file format that tools and ac
 
 - `key_id`
 - `issuer_id`
-- `key` (the content of the `.p8` file)
+- `key_content` (the content of the `.p8` file)
 
 The JSON file allows optional:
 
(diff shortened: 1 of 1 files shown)


rogerluan:
@garrettmoon thanks for your PR. I initially thought you were right, but after investigating a bit, looks like `key` is the correct _key_. That JSON file is read here:

https://github.com/fastlane/fastlane/blob/77f0e1ad15aac123f3d2100a84fb18bc5a6ef16d/spaceship/lib/spaceship/connect_api/token.rb#L34

And there, you can see that the key being looked up is really "key" and not "key_content" (which is different from the [app_store_connect_api_key action](https://docs.fastlane.tools/actions/app_store_connect_api_key/) `key_content`) 

Closing this PR 😊 Let me know if you have further questions!

garrettmoon:
I think this *was* correct and the docs have already been updated to reflect that here (no more reference to `key`): https://docs.fastlane.tools/actions/app_store_connect_api_key/

rogerluan:
@garrettmoon the changes you made in this PR aren't related to the link you added to the PR description (which's the same link you're posting now) 😅 they're 2 different pages. The changes you made in the PR refer to this page instead: https://docs.fastlane.tools/app-store-connect-api/

garrettmoon:
🤦 thank you for taking the time to not only look at my PR but kindly explain the mistake I was making!

rogerluan:
Of course @garrettmoon ! It's the beauty of community-driven projects ❤️

PR #995, opened 2020-10-21
Fix link of guide "distributing beta builds for React Native projects"

AurelGit:
Same documentation but has just moved to a new url


rogerluan (approved):
Thanks for fixing this!

rogerluan (approved):

PR #1013, opened 2020-12-28
Update installing-fastlane.md

artemisia-absynthium:
Update install documentation to reflect breaking changes in Xcode.

Now headers are not included anymore in the MacOS provided Ruby installation, so dependencies that need to build native extensions would throw an error and make the install fail. Now all the Ruby related commands need to pass through Xcode via `xcrun` (see comment in issue https://github.com/fastlane/fastlane/issues/15183#issuecomment-543844264).
I think this way is cleaner than installing another version of Ruby that could cause conflict with the preinstalled one.

Talking about cleanliness, I included an extra optional step to install dependencies in the project folder instead of system wide, to avoid the need to use sudo and to avoid cluttering system folders. I personally prefer this approach for cleanliness and safety both, so I think it's nice to tell there's this option too.

I came to this conclusions after some hours of research on the web and reading various docs, btw any further advice is welcome!

### DIFF
--- a/docs/includes/installing-fastlane.md
+++ b/docs/includes/installing-fastlane.md
@@ -10,15 +10,28 @@ source "https://rubygems.org"
 
 gem "fastlane"
 ```
-- Run `[sudo] bundle update` and add both the `./Gemfile` and the `./Gemfile.lock` to version control
+- If you want the gems installed in a folder inside your project instead of globally in the system, so that you can avoid the need for `sudo`, you can run `bundle config set --local path 'vendor/bundle'`
+- Run
+  - Linux/Windows: `[sudo] bundle update`
+  - macOS: `[sudo] xcrun bundle update`
+- Add both the `./Gemfile` and the `./Gemfile.lock` to version control
 - Every time you run _fastlane_, use `bundle exec fastlane [lane]`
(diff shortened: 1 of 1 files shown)


rogerluan (changes requested):
Hey @artemisia-absynthium 👋 

Now that https://github.com/fastlane/docs/pull/1022 was merged, is this PR still relevant? If so, would you mind rebasing this PR and updating the docs accordingly? 🙏 😊

rogerluan:
On a second review:

>I think this way is cleaner than installing another version of Ruby that could cause conflict with the preinstalled one.

This is actually not true 🙈 Using the native ruby is not recommended, hence why we added to our docs that one should prefer to use a ruby version manager 😊 Ruby (and other scripting languages) actually won't be included in macOS by default in future versions of macOS [according to Apple](https://developer.apple.com/documentation/macos-release-notes/macos-catalina-10_15-release-notes).

We appreciate your contribution, but I'll close this PR since its content has been completely revamped already in master since this was open. Let me know if you have further questions! 🤗

artemisia-absynthium:
Hi @rogerluan, sorry for the late response and thank you very much for pointing me to that important information! This changes everything! 😄

PR #1019, opened 2021-02-03
[spaceship] document usage of SPACESHIP_ONLY_ALLOW_INTERACTIVE_2FA env var

kohtenko:
Update docs for https://github.com/fastlane/fastlane/pull/18109

This PR shouldn't be merged unless https://github.com/fastlane/fastlane/pull/18109 is approved and merged 🙏

### DIFF
--- a/docs/best-practices/continuous-integration.md
+++ b/docs/best-practices/continuous-integration.md
@@ -87,7 +87,7 @@ Unfortunately there is nothing _fastlane_ can do better in this regard, as these
 ## Environment variables to set
 
 Most setups will need the following environment variables
-
+- `SPACESHIP_ALLOW_NON_INTERACTIVE_2FA`: Explicitly allow to run 2FA in non-interactive mode.
 - `FASTLANE_USER`: Your App Store Connect / Apple Developer Portal user, if your _fastlane_ setup accesses App Store Connect or the Apple Developer Portal (e.g. submit a TestFlight build, create a profile, ...)
 - `FASTLANE_PASSWORD`: Your App Store Connect / Apple Developer Portal password, usually only needed if you also set the `FASTLANE_USER` variable
 - `MATCH_PASSWORD`: You need to provide the password of your _match_ encryption if you use _match_
--- a/docs/best-practices/continuous-integration.md
+++ b/docs/best-practices/continuous-integration.md
@@ -87,7 +87,7 @@ Unfortunately there is nothing _fastlane_ can do better in this regard, as these
 ## Environment variables to set
 
 Most setups will need the following environment variables
-- `SPACESHIP_ALLOW_NON_INTERACTIVE_2FA`: Explicitly allow to run 2FA in non-interactive mode.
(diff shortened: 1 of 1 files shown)


rogerluan (changes requested):
This variable should be moved to the bottom of the list, as it's usually less important than e.g. `FASTLANE_USER`.

rogerluan (approved):
I'm approving but I'm not sure the language is clear enough. I'm not a native English speaker, but I think it's confusing the way it's phrased 😬 

Maybe @janpio or @ainame can take a look here? 🤗 

Other than that, LGTM! Thanks for making this change!

ainame (approved):
> I'm approving but I'm not sure the language is clear enough. I'm not a native English speaker, but I think it's confusing the way it's phrased 😬

I'm not a native English speaker either😂 but in general, I think docs should also be understandable for non-native speakers. I know what @rogerluan felt. It feels like something similar to a double negative.

However, from my non-native perspective, in this case, people read the variable name first, and its meaning is quite clear already (at least to me) so I'm fine with this as the description just says in the other way around to clarify 🙂

rogerluan:
Yeah I think the double negative (`prevent` + `non`-interactive) is what makes it confusing 😅  not sure how I'd improve it, though. Can't think of a way out lol

docs/best-practices/continuous-integration.md
```diff
@@ -94,6 +94,7 @@ Most setups will need the following environment variables
 - `FASTLANE_APPLE_APPLICATION_SPECIFIC_PASSWORD`: You need to provide an [application specific password](#application-specific-passwords) if you have 2-factor enabled and use _pilot_ or _deliver_ to upload a binary to App Store Connect
 - `FASTLANE_SESSION`: You need to provide a [pregenerated session via `fastlane spaceauth`](#spaceauth) if you have 2-factor authentication enabled and want to use any actions that communicates with App Store Connect.
 - `LANG` and `LC_ALL`: These set up the locale your shell and all the commands you execute run at. _fastlane_ needs these to be set to an UTF-8 locale to work correctly, for example `en_US.UTF-8`. Many CI systems come with a locale that is unset or set to ASCII by default, so make sure to double-check whether yours is set correctly.
+- `SPACESHIP_ONLY_ALLOW_INTERACTIVE_2FA`: Explicitly prevent running of 2FA in non-interactive mode.
```
rogerluan, line 97:
```suggestion
- `SPACESHIP_ONLY_ALLOW_INTERACTIVE_2FA`: Explicitly prevents being prompted for 2FA in non-interactive mode.
```

@kohtenko @ainame is this more clear and still make sense to what the env var does? I think it reads better

ainame: Yeah it does but I come to think this env var is the workaround that we should ideally fix on the code side😂

rogerluan: I thought the env var was added _as the fix_ 😅 ?

rogerluan (approved):

PR #1029, opened 2021-03-06
document the fact that App Store Connect API sessions max duration is 20 minutes

simonbengtsson:
Clarify that maximum session length is 1200 seconds (20 min) as mentioned in the app store api docs https://developer.apple.com/documentation/appstoreconnectapi/generating_tokens_for_api_requests

If set to a higher number fastlane will output the following error. It would be great with a code check for this as well with a more friendly error message.

```
[!] The request could not be completed because:
        Authentication credentials are missing or invalid. - Provide a properly configured and signed bearer token, and make sure that it has not expired. Learn more about Generating Tokens for API Requests https://developer.apple.com/go/?id=api-generating-tokens
```

### DIFF
--- a/docs/app-store-connect-api.md
+++ b/docs/app-store-connect-api.md
@@ -55,7 +55,7 @@ lane :release do
     key_id: "D383SF739",
     issuer_id: "6053b7fe-68a8-4acb-89be-165aa6465141",
     key_filepath: "./AuthKey_D383SF739.p8",
-    duration: 1200, # optional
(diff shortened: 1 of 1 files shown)


simonbengtsson:
@googlebot I signed it!

rogerluan (approved):
Oh, this is a great find! Thanks @simonbengtsson !! 🚀 

I'll make sure to open a PR with this validation in _fastlane_ 💪

PR #1043, opened 2021-03-27
Fix Typo in setup.md

Bluzkry:
In the current documentation, we have the below. https://docs.fastlane.tools/getting-started/android/setup/#collect-your-google-credentials
-- --

```
json_key_file("path/to/your/play-store-credentials.json")
package_name("my.package.name")


The path is relative to where you normally run `fastlane`.

### Configure _supply_

Edit your `fastlane/Appfile` and change the `json_key_file` line to have the path to your credentials file:

```ruby
json_key_file "/path/to/your/downloaded/key.json"
```
.
.
.
I think that the **Google Credentials** and **Configure _supply_** steps are mixed together when they should be separate steps. I believe that this should be changed to something like the below.
-- --
```
json_key_file("path/to/your/play-store-credentials.json")
package_name("my.package.name")
```

The path is relative to where you normally run `fastlane`.

### Configure _supply_

Edit your `fastlane/Appfile` and change the `json_key_file` line to have the path to your credentials file:

```
ruby
json_key_file "/path/to/your/downloaded/key.json"
```

### DIFF
--- a/docs/includes/google-credentials.md
+++ b/docs/includes/google-credentials.md
@@ -23,5 +23,5 @@ You can use [`fastlane run validate_play_store_json_key json_key:/path/to/your/d
 ```ruby
 json_key_file("path/to/your/play-store-credentials.json")
 package_name("my.package.name")
-
+```
 The path is relative to where you normally run `fastlane`.


janpio (approved):

janpio:
Good catch, thanks for reporting and opening a PR even 👍

PR #1051, opened 2021-04-16
Update mkdocs from 0.15.3 to 1.1.2

joshdholtz:
Super old

PR #1085, opened 2021-08-10
Fix sidenavv.css 404 not found

janbrasna:
Removes missing/non-existent extra css:

> Screen Shot 2021-08-10 at 14 32 55

It was introduced with https://github.com/fastlane/docs/pull/1051/commits/d87a7c08d2dfb051df7b073d26629ee6790625c1 when migrating `mkdocs` in #1051 — seems like an error or some leftover from testing different cascade ordering to me.

I did my best to go through the whole PR branch (dope commit messages 👍) to trace it back to more context around the change, and can't find any other reasoning. No such file is in the source, nothing like that is being built to `gh-pages`, and all netlify builds since also display this 404 error. Which leads me to thinking this is a mere typo or omission, not a shim for testing or something undocumented, made on purpose that should've made it into prod.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -21,7 +21,6 @@ markdown_extensions:
     toc_depth: 2
 docs_dir: docs
 extra_css:
-- "./css/sidenavv.css"
 - "./css/fastlane.css"
 - "./css/code.css"
 - "./css/breadcrumbs.css"


janbrasna:
BTW there's also a plethora of new files added at that point in 044d27f845fe8a06c703f27c71fdbe53092e1dd2 at `/theme` as:
- this_toc.html
- working_toc.html
- xbreadcrumbs.html
- xtoc.html

that all sound like a temporary stage that got into master — trying to find any use, with no luck. Just nod, and I'll prune them as well.

janbrasna:
@joshdholtz Aww no worries, been an honor;) 😇 Thanks for taking the plunge and spending a good part of this summer reconstructing pretty much everything for the upgrade to work;)

I just wanted to confirm these random `*_toc` and `*_breadcrumbs` files are indeed some testing leftovers, and are not needed in some edge cases I wasn't able to infer. — It's been like that in prod for about a month, so I guess it's safe & they won't be missed at all;))

It's not about mess _per se_. I tripped over them when working on an unrelated PR/fix. So curious about their origins, I started to dig deeper in the upgrade history trying to make my way around the repo changes at the time (and possible errors, not knowing if they were introduced by my edits or been there already pre-existent…). Therefore this PR proposal was merely an attempt to save the same head scratching learning curve for any newcomer contributor, trying to wrap their heads around the same questions about the app structure and context/usage as I experienced. Just trying to make the repo more welcoming for noobs like myself;)

PR #1090, opened 2021-08-26
add information about API Key appropriate role selection

### LINKED ISSUE #1050: Mention "App Manager" role on App Store Connect API docs
http://docs.fastlane.tools/app-store-connect-api/

After reading this page, I created an API key in https://appstoreconnect.apple.com/access/api as instructed, and of the five available access levels (Admin, App Manager, Developer, Finance, Sales and Reports) I chose "Developer".

But it turns out that in the `upload_to_testflight` action, setting `changelog` with a "Developer" API key active results in the following error:
`Could not set changelog: This request is forbidden for security reasons - The API key in use does not allow this request`

and setting `beta_app_feedback_email` or `beta_app_description` results in:
`Could not set beta_app_feedback_email and/or beta_app_description: This request is forbidden for security reasons - The API key in use does not allow this request`

After trying again with an "App Manager" API key, no such problems.

Could we update the docs to mention that "App Manager"-level permissions are needed for some fields, including these three?

torgefaehrlich:
Should close #1050

### DIFF
--- a/docs/app-store-connect-api.md
+++ b/docs/app-store-connect-api.md
@@ -34,7 +34,7 @@ Below are the statuses of each tool:
 
 1. Create a new App Store Connect API Key in the [Users page](https://appstoreconnect.apple.com/access/api)
   - For more info, go to the [App Store Connect API Docs](https://developer.apple.com/documentation/appstoreconnectapi/creating_api_keys_for_app_store_connect_api)
-  - Take care to choose a role appropriate for the task at hand from [Roles for the Apple Developer Program](https://developer.apple.com/support/roles/)
+  - Give your API Key an appropriate role for the task at hand. You can read more about roles in [Permissions in App Store Connect](https://developer.apple.com/support/roles/)
 2. Download the newly created API Key file (`.p8`)
   - This file cannot be downloaded again after the page has been refreshed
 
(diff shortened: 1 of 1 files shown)


docs/app-store-connect-api.md
```diff
@@ -34,6 +34,7 @@ Below are the statuses of each tool:
 
 1. Create a new App Store Connect API Key in the [Users page](https://appstoreconnect.apple.com/access/api)
   - For more info, go to the [App Store Connect API Docs](https://developer.apple.com/documentation/appstoreconnectapi/creating_api_keys_for_app_store_connect_api)
+  - Take care to choose a role appropriate for the task at hand from [Roles for the Apple Developer Program](https://developer.apple.com/support/roles/)
```
rogerluan, line 37:
```suggestion
  - Give your API Key an appropriate role for the task at hand. You can read more about roles in [Permissions in App Store Connect](https://developer.apple.com/support/roles/)
```

rogerluan (changes requested):
Thanks for taking care of this issue! 🤗 I agree with your suggestion to improve the docs. I've just improved wording a little bit. If you could apply the suggestions I made, this is good to be merged in 💪

rogerluan (approved):
Thank you for your contribution 🚀  and for addressing the changes requested so promptly 😊

PR #1075, opened 2021-07-15
Fix markdown not parsed in foldable details

lucgrabowski:
### Motivation and Context
Resolves #1074
Markdown is not parsed in foldable details html tag, e.g. on https://docs.fastlane.tools/actions/deliver/

### Description
Added  `markdown.extensions.md_in_html` extension to parse markdown in html tags.
Used `markdown="1"` to parse markdown.

Fix based on:
https://github.com/mkdocs/mkdocs/issues/1687
https://python-markdown.github.io/extensions/md_in_html/

### Testing Steps
Install mkdocs locally, build site, check html pages.
Dear reviewer, please check if the changes in formatting are fine.

PR #1091, opened 2021-08-30
Group auto-generated files in one directory

lucgrabowski:
### Motivation and context
`actions.md`, actions documentation and `available-plugins.md` are moved into `/generated` directory as proposed in https://github.com/fastlane/docs/pull/1075#issuecomment-895661648

**This PR should be considered together with:** https://github.com/fastlane/fastlane/pull/19309
 

### Description

Urls in actions documentation will change from:
- actions to generated/actions/<ACTION>, e.g.:
https://docs.fastlane.tools/actions/deliver to https://docs.fastlane.tools/generated/actions/deliver
- https://docs.fastlane.tools/actions/ to https://docs.fastlane.tools/generated/actions/
- https://docs.fastlane.tools/plugins/available-plugins to https://docs.fastlane.tools/generated/available-plugins/

Changes:
- Moved `actions.md` to `generated` directory
- Moved actions documentation to `generated/actions` directory
- Moved `plugins/available-plugins.md` to `generated` directory
- Updated `mkdocs.yml` - also actions list that will be regenerated by fastlane's `markdown_docs_generator`
- Updated documentation with  `bundle exec fastlane update_docs debug:true skip_plugin_scores:true` (https://github.com/fastlane/fastlane/blob/9b6d47e74da8cf05a8993ec8b4c2c66a466ce358/fastlane/Fastfile#L400)
- Updated theme (breadcrumbs)
- Removed `edit on github` button from available plugins docs
- Updated `available-redirects`
- Updated deployment script

### Testing Steps

Install mkdocs locally, build site, check html pages. Check links on pages.

I did initial testing of links. I couldn't test deployment.


janpio (changes requested):
Nice!

> Urls in actions documentation will change from:

This unfortunately is a no go with the existing URLs are not also redirected to these new URLs at the same time. This would break many, many links all over the internet from the last x years.

Even with redirects in place, this solution would not be optimal. Much better would be if the URL would not change in the first place. Is there a way to achieve that?

lucgrabowski:
@janpio Thanks for review. I totally agree, too many links would have been broken that way.
Unfortunately I don't see an easy way to move pages to subdirectory without changing URLs, mkdocs creates those urls automatically.
To make redirects work, we could use `mkdocs-redirect` plugin to redirect to `generated` directory. It is not the best, but it's possible to achieve it easily. I've pushed a commit to show how it can be done.

janpio:
> mkdocs creates those urls automatically.

I think this is the problem here - if the URLs are 100% tied to the file location, then it will be hard to have the files in `generated` but keep the old URLs.

Idea: Could the files be moved to the correct location (without `generated`) during the build to create the old URLs?

lucgrabowski:
I assume that copying generated files could be done in `deploy.sh` script before `mkdocs build` but in such case, I'm not really sure if vaildation from `Fastfile` would work.

janpio:
Certainly not out of the box. The idea would be to have the file in the repo in a structure that makes it more obvious what is manual and what is generated, but put it all together during/before build so the output is still the same as now.

lucgrabowski:
I've opened a new PR for copying generated files into proper locations: https://github.com/fastlane/docs/pull/1095

lucgrabowski:
I've opened a new PR for copying generated files into proper locations: https://github.com/fastlane/docs/pull/1095

PR #1099, opened 2021-09-06
Remove "Edit on Github" button from documentation of actions

lucgrabowski:
"Edit on Github" button should not be present on generated actions documentation pages.
Right now following the link gives 404 error.

Should `xbreadcrumbs.html` be changed in a similar way?

### DIFF
--- a/theme/breadcrumbs.html
+++ b/theme/breadcrumbs.html
@@ -13,7 +13,9 @@
     {% if page %}<li>{{ page.title }}</li>{% endif %}
 
 
-    {% if page and page.file.src_path != "actions.md" and not "actions/" in page.file.src_path and page.file.src_path != "plugins/available-plugins.md" %}
+    {% if page and page.file.src_path != "actions.md" 
+          and not "actions/" in page.file.src_path 
+          and page.file.src_path != "plugins/available-plugins.md" %}
       <li class="wy-breadcrumbs-aside">
         {%- block repo %}
         {% if page and page.edit_url %}
--- a/theme/breadcrumbs.html
+++ b/theme/breadcrumbs.html
@@ -13,7 +13,7 @@
     {% if page %}<li>{{ page.title }}</li>{% endif %}
 
 
-    {% if page and page.file.src_path != "actions.md" and page.file.src_path != "plugins/available-plugins.md" %}
+    {% if page and page.file.src_path != "actions.md" and not "actions/" in page.file.src_path and page.file.src_path != "plugins/available-plugins.md" %}
       <li class="wy-breadcrumbs-aside">
         {%- block repo %}
         {% if page and page.edit_url %}


rogerluan:
Which page specifically are you referring to? 

http://docs.fastlane.tools/actions/ - this one shows no "Edit on GitHub" button for me
http://docs.fastlane.tools/advanced/actions/ - this one does, but it should show cuz it's not a generated page

janpio:
I assume it is about pages like https://docs.fastlane.tools/actions/run_tests/ (the link is currently broken as the pages got moved to a subfolder - but GH suggests that correctly for me)

The value linking to edit these files has is this comment at the top of the page:
```
<!--
This file is auto-generated and will be re-generated every time the docs are updated.
To modify it, go to its source at https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/run_tests.rb
-->
```

This makes it discoverable where you have to go to edit something.

Alternatively we could of course link there directly. I am not sure how feasible this is though.

lucgrabowski:
Sorry for not making it clear enough. It's about buttons on action details pages as @janpio said.
Example of a broken link from https://docs.fastlane.tools/actions/run_tests/ :
https://github.com/fastlane/docs/edit/master/docs/actions/run_tests.md

Comments at the top of the page are generated from template:
https://github.com/fastlane/fastlane/blob/master/fastlane/lib/assets/ActionDetails.md.erb
and linking always to `.rb` files.
https://github.com/fastlane/fastlane/blob/f2aa10614bd41025e3b463ad3754126b16bb2585/fastlane/lib/assets/ActionDetails.md.erb#L1-L4
E.g.:
```
<!--
This file is auto-generated and will be re-generated every time the docs are updated.
To modify it, go to its source at https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/run_tests.rb
-->
```

Removing the buttons from action detail pages would make them consistent with `plugins/available_pages.md` and `actions.md` that are shown without edit button.
If we would like to leave the button, I'm not really sure how to generate links without hardcoding them to fastlane repository. For now current page edit url is used: `page.edit_url`
https://github.com/fastlane/docs/blob/0c051ed7ff6f585cd417fef9900b6df61651b6ec/theme/breadcrumbs.html#L16-L19

rogerluan:
I don't have a strong preference moving forward. I think removing the button would be like a hotfix, but ideally we should take the users _somewhere_. We know that those pages are generated and that they'd be overridden on re-generation of this repo, but if we gave users an easy way to interact with the docs and provide patches, it'd be easier for us to notice it and figure out where is the appropriate place to fix that, and go ahead and fix it (e.g. updating some doc comment in fastlane/fastlane repo).

janpio:
As anecdata: I did use the button all the time when I still regularly updated the docs of actions. Otherwise I often would not have bothered, but via this path I could get to the action source code reasonably fast. A direct button click to the action would be much better though.

lucgrabowski:
I was able to generate valid urls and opened a follow up PR (based on this one): https://github.com/fastlane/docs/pull/1100

lucgrabowski:
Hi @joshdholtz, would you be able to take a look at this PR and a follow up PR? Edit buttons were broken for generated docs. I've removed edit buttons in this PR and proposed a way of restoring and fixing them in the second one.

janbrasna:
Hey @lucgrabowski! Sorry to bother you but I have one followup question… You mentioned:

> Should xbreadcrumbs.html be changed in a similar way?

Have you been able to find out if that file is used in any way? I've removed it along with others in https://github.com/fastlane/docs/pull/1085#issuecomment-896086752 and it seems working since… however I've never really truly found out what it ever was for;) So maybe you had better insight into what versions are used where? Thanks!

lucgrabowski:
@janbrasna Hi, I think that `xbreadcrumbs.html` was one of the files that weren't used at all. I see that you cleaned some of the files a month ago. I don't have insight on it ;)

PR #1100, opened 2021-09-08
"Edit on Github" button for action details opens source file in fastlane repository

lucgrabowski:
"Edit on Github" button for action details is restored.
Links for source `.rb` files are created based on
- `fastlane_repo_base_edit_uri` from `mkdocs.yml` and 
- `page.file.src_path` with changed extension from `.md` to .rb`

### DIFF
--- a/theme/breadcrumbs.html
+++ b/theme/breadcrumbs.html
@@ -13,7 +13,9 @@
     {% if page %}<li>{{ page.title }}</li>{% endif %}
 
 
-    {% if page and page.file.src_path != "actions.md" and not "actions/" in page.file.src_path and page.file.src_path != "plugins/available-plugins.md" %}
+    {% if page and page.file.src_path != "actions.md" 
+          and not "actions/" in page.file.src_path 
+          and page.file.src_path != "plugins/available-plugins.md" %}
       <li class="wy-breadcrumbs-aside">
         {%- block repo %}
         {% if page and page.edit_url %}
--- a/theme/breadcrumbs.html
+++ b/theme/breadcrumbs.html
@@ -12,22 +12,27 @@
     {% endif %}
     {% if page %}<li>{{ page.title }}</li>{% endif %}
 
-
-    {% if page and page.file.src_path != "actions.md" 
-          and not "actions/" in page.file.src_path 
-          and page.file.src_path != "plugins/available-plugins.md" %}
+    {% if page and page.file.src_path != "actions.md" and page.file.src_path != "plugins/available-plugins.md" %}
       <li class="wy-breadcrumbs-aside">
         {%- block repo %}
-        {% if page and page.edit_url %}
-          <a href="{{ page.edit_url }}"
+
+        {% set icon_class_block %}
           {%- if config.repo_name|lower == 'github' %}
             class="icon icon-github"
           {%- elif config.repo_name|lower == 'bitbucket' %}
             class="icon icon-bitbucket"
           {%- elif config.repo_name|lower == 'gitlab' %}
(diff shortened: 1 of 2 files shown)


lucgrabowski:
Validation of mkdocs configuration failed even though page could be built without issues.
Running `mkdocs build --verbose --clean --strict` showed in logs:
```
WARNING -  Config value: 'fastlane_repo_base_edit_uri'. Warning: Unrecognised configuration name: fastlane_repo_base_edit_uri 
...
DEBUG   -  Config value: 'fastlane_repo_base_edit_uri' = 'https://github.com/fastlane/fastlane/edit/master/fastlane/lib/fastlane' 
```
It seems that custom values should be in `extra` dictionary in `mkdocs.yml`

lucgrabowski:
I've tested links and noticed that there are 6 that probably didn't work for some time and are still broken after changes from this PR:

1.  from xcodebuild.rb (https://github.com/fastlane/fastlane/edit/master/fastlane/lib/fastlane/actions/xcodebuild.rb): 
- xctest
- xcclean
- xcarchive
- xcexport
- xcbuild
2. plugin_scores (should be https://github.com/fastlane/fastlane/edit/master/fastlane/actions/plugin_scores.rb)

theme/breadcrumbs.html
```diff
@@ -12,22 +12,30 @@
...
+
+        {% set edit_docs_file_url %}
+          {%- if "actions/" in page.file.src_path %}
+            {{ config.extra.fastlane_repo_base_edit_uri }}/{{ page.file.src_path|string|safe|replace(".md", ".rb") }}
+          {%- elif page.edit_url %}
+            {{ page.edit_url }}
+          {%- endif %}
+        {% endset %}
+
+        <a href="{{ edit_docs_file_url }}" {{ icon_class_block }}>Edit on {{ config.repo_name }}</a>
```
janbrasna, line 37: I'm struggling with the whitespace around all the `% set` blocks… What's the best way around here? For source legibility this is okay. However it produces slightly wild output (including whitespace in the actual anchor link… that luckily gets stripped in all the UAs I've tried… but is not really the intended output I guess):

Screen Shot 2021-11-21 at 1 04 41

Would it be OK to somehow inline the code blocks a little more, so the template parser doesn't inject all the whitespace into the variable content?

theme/breadcrumbs.html
```diff
@@ -26,12 +26,15 @@
...
-        Edit on {{ config.repo_name }}</a>
+        {% set edit_docs_file_url %}
+          {%- if "actions/" in page.file.src_path %}
+            {{ config.fastlane_repo_base_edit_uri }}/{{ page.file.src_path|string|safe|replace(".md", ".rb") }}
+          {%- elif page.edit_url %}
+            {{ page.edit_url }}
+          {%- endif %}
+        {% endset %}
+
+        <a href="{{ edit_docs_file_url }}" {{ icon_class_block }}>Edit on {{ config.repo_name }}</a>
```
janbrasna, line 37: This would be better:

```suggestion
        <a href="{{ edit_docs_file_url }}" {{ icon_class_block }}> Edit on {{ config.repo_name }}</a>
```

Screen Shot 2021-11-21 at 2 21 33

I know, it's weird from a semantic point of view to have the space an integral part of the anchor text… however that's how the original sphinx/rtd template does it to prevent exactly this aligment issue. 🤷

janbrasna (commented):
I was about to propose a couple of tiny fixes only to realise they lead here, so trying to get some feedback how to go about them before trying addressing these in a PR…

PR #1101, opened 2021-09-13
Improve appstore connect api key docs

lucgrabowski:
Motivation:
https://github.com/fastlane/fastlane/issues/19348
https://github.com/fastlane/fastlane/issues/19341

Changes:
Rename `key_content` to `key` in JSON file format, as expected in `Token`
https://github.com/fastlane/fastlane/blob/be624617eb2cefb5a96457f27768ad26d74eebed/spaceship/lib/spaceship/connect_api/token.rb#L52-L62
Explain how to use `api_key` parameter in actions and tools

### DIFF
--- a/docs/app-store-connect-api.md
+++ b/docs/app-store-connect-api.md
@@ -85,7 +85,10 @@ end
 
 ### Using _fastlane_ API Key hash option
 
-Expected keys and values are described in _fastlane_ API Key JSON file format below
+Keys and values that can be used in hash in `api_key` parameter in actions (e.g. `upload_to_testflight`) and tools (e.g. `cert`) are described in _fastlane_ API Key JSON file format below.
+The only difference when using hash is that you could use `filepath` instead of `key`.
+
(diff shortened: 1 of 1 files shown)


rogerluan:
I'm confused - what about https://github.com/fastlane/docs/issues/1079 and its associated PR? That's what caused this regression in the docs, but _fastlane_ seems to expect `key_content`? Does it look for `key_content` in some calls sites but not in others? 🤔 @lucgrabowski

rogerluan:
I do believe your suggested changes are the correct form, given the tickets that it aims to solve. But I'd like to understand #1079 first before merging this in 😬

joshdholtz (approved):
This looks like a great change! I actually had no idea I made the `key` and `key_content` different for each approach 😬 

I'm down for merging this but I think I'm doing to made add support for allowing `key_content" in the JSON format so that it can be the same 🤷‍♂️

joshdholtz:
@rogerluan The issue is the action looks for `key_content` but the loading from file uses `key` which his very confusing 😔 Not sure why I did that 🤦‍♂️

PR #696, opened 2018-08-21
[WIP] Change site theme to Material

revolter:
[Preview](https://deploy-preview-696--fastlane-docs-preview.netlify.com/)

- [x] Resolves #689 
- [ ] Fix #619
- [ ] Make sure `wy-breadcrumbs` still exists [for Algolia](https://github.com/algolia/docsearch-configs/blob/71041536cc0fa8255a84c43788b99be81760dcba/configs/fastlane.json#L12)

PR #1107, opened 2021-10-05
Separate mkdocs.yml content into multiple sections

revolter:
The only useful thing that I could salvage from #696 😂


rogerluan (approved):
😁 Nice!

PR #1112, opened 2021-10-15
Clarified Xcode project configuration for code signing

erkhembayar-gantulga:
Xcode 8 section was a bit confusing because it isn't specific for
Xcode 8, but actually it can be applicable for Xcode 8 and up.
Therefore, changed the section title in order to clarify it.

Note:
We used to `build` release from XCode  directly with `Automatic Code Signing`. So, I tried to configure `fastlane`with my existing project. So, I followed the doc instruction but I get stuck at [codesigning](https://docs.fastlane.tools/codesigning/getting-started/) for 2 days. Because I skipped `XCode 8` section immediately every time I come to `Setting up your Xcode Project` page

### DIFF
--- a/docs/codesigning/xcode-project.md
+++ b/docs/codesigning/xcode-project.md
@@ -4,7 +4,7 @@
 
 # Xcode 9 and up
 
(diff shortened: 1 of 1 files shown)


erkhem42:
@googlebot I signed it

PR #1113, opened 2021-10-21
Add ASC API status for `download_dsyms` action, and mark `pem` as available via Apple ID auth

### LINKED ISSUE #1111: App Store Connect API documentation missing download_dsyms
Documentation for [Using App Store Connect API](https://docs.fastlane.tools/app-store-connect-api/) does not include info on [download_dsyms](https://docs.fastlane.tools/actions/download_dsyms/), which makes the usage misleading.

If `pem` is included as `No/No` under **Supported Actions/Tools**, I would have expected to see `download_dsyms` with the same information. 

I would have opened this as a Pull Request, but I assumed there may be a reason why and wanted to raise as an issue first. I'm also not sure if there are other App Store Connect-related actions that could be included in this list for completion.

Thank you!

revolter:
Fixes #1111.

### DIFF
--- a/docs/app-store-connect-api.md
+++ b/docs/app-store-connect-api.md
@@ -29,6 +29,7 @@ Below are the statuses of each tool:
 | produce | **Partial** | No |
 | pem | **Yes** | No |
 | precheck | **Yes (except for IAP)** | **Yes (except for IAP)** |
+| download_dsyms | **Yes** | No |
 
 ## Creating an App Store Connect API Key
 


bdrelling:
@rogerluan thank you for the quick turn-around time on this one! Sorry I didn't open a PR myself, I haven't had much time lately and wasn't confident I'd be able to see it through.

revolter:
Duuude, I can't believe the coincidence:

https://github.com/fastlane/fastlane/releases/tag/2.197.0

So I guess that a new PR is needed 😆 @bdrelling, do you want to do it?

PR #1121, opened 2021-11-20
Set Python version to 3.7 in Netlify

revolter:

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,2 @@
+[build.environment]
+  PYTHON_VERSION = "3.7"


revolter:
Still doesn't work 😡

janbrasna:
@revolter I understand that’s attempting to fix the netlify builds missing weakref? It seems getting rid of the backports might help, however I have clearly no idea where the env (docker config maybe?) lives, as there’s no reference to weakref in this repo.

Context: https://github.com/pypa/pipenv/issues/4834

revolter:
@janbrasna, I think that it's [a Netlify issue](https://answers.netlify.com/t/issues-with-netlify-python-requests/48018/), and they're the one that should fix it. It broke when they updated the docker image that they're using.

janbrasna:
@revolter Gotcha. That's why I couldn't find any of that in the tooling/pipeline;D Good thing is it fixed itself at some point later in December, so one can rely on previews again. 🚀

janbrasna:
@revolter So I guess it's time to try this again, as the currently used 3.5 in the image does pretty much nothing aside from throwing a jillion of EOL warnings;D — according to https://github.com/netlify/build-image/blob/xenial/included_software.md#languages this value should work. (Can't understand why they default to EOLed version as a the right choice for python3 instead of the highest they support, beyond me:/…)

Otherwise I think it means either specifying exact minor 3.7 in Pipfile _(which makes things unnecessarily specific for local work?)_ or providing runtime.txt with 3.7 _(pretty much tidying up these 044d27f + 4240507)_?

janbrasna:
Yea I can only confirm that the `.toml` file had no impact during my experimenting in #1164 … no idea why:/

PR #1126, opened 2022-01-01
Update mkdocs copyright from 2021 to 2022

Kukurijek:
The copyright date displayed in the documentation has been updated.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -3,7 +3,7 @@ site_name: fastlane docs
 site_author: fastlane team
 site_description: Documentation for fastlane tools, the easiest way to automate building
   and releasing your iOS and Android apps
-copyright: Copyright &copy; 2021 <a href="https://fastlane.tools">fastlane</a>
+copyright: Copyright &copy; 2022 <a href="https://fastlane.tools">fastlane</a>
 repo_url: https://github.com/fastlane/docs
 theme:
   name: readthedocs


rogerluan (approved):
Happy new year! 😁

rogerluan:
@revolter @janpio do you know if there's a better way to do this so that we don't have to update this every year? Something like a dynamic date? 😄

Kukurijek:
> @revolter @janpio do you know if there's a better way to do this so that we don't have to update this every year? Something like a dynamic date? 😄
> I don't mind this easy change once a year, but jw if you remember something from the top of your head :)

Most likely the utc datetime object `build_date_utc` from [mkdocs](https://www.mkdocs.org/dev-guide/themes/#build_date_utc) can be used to display the current year 😄

rogerluan:
@Kukurijek do you think you could make this change here instead of hardcoding the "2022"? That'd be super 😁😁

revolter:
I think that that variable can be used in the HTML template file, not in the YAML config file itself:

https://github.com/zendframework/zf-mkdoc-theme/commit/597f50335f80a34cfa3270c2fada9a64bbcbc896

Which means that we would need to stop using the `copyright` config item, and put the HTML snippet directly in the template.

Kukurijek:
> I think that that variable can be used in the HTML template file, not in the YAML config file itself:
> 
> [zendframework/zf-mkdoc-theme@597f503](https://github.com/zendframework/zf-mkdoc-theme/commit/597f50335f80a34cfa3270c2fada9a64bbcbc896)
> 
> Which means that we would need to stop using the `copyright` config item, and put the HTML snippet directly in the template.

That is correct. Second option is to use '`r format(Sys.Date(), "%B %d, %Y")`' instead of "2022" ->
https://stackoverflow.com/questions/23449319/yaml-current-date-in-rmarkdown

(I tested both and they are both functioning)
What do you consider to be a better approach?

revolter:
> That is correct. Second option is to use '`r format(Sys.Date(), "%B %d, %Y")`' instead of "2022" -> [stackoverflow.com/questions/23449319/yaml-current-date-in-rmarkdown](https://stackoverflow.com/questions/23449319/yaml-current-date-in-rmarkdown)
> 
> (I tested both and they are both functioning) What do you consider to be a better approach?

Interesting idea 🤔 But we're not using `rmarkdown`. I also tested it, and it doesn't work.

I think that it works for you because you have `rmarkdown` or `R` installed in the system, and they're somehow automatically used.

So, I'd say to move the copyright snippet in the HTML template, instead of adding a new dependency just for this small detail. @rogerluan, what's your take? 😃

rogerluan:
I agree @revolter ! The less dependencies, the better 😊 I'm also in favor of modifying the HTML template

Kukurijek:
I created a new PR -> #1128
So feel free to close this one.
Thanks for the feedback! 🚀

rogerluan:
Thanks @Kukurijek !

PR #1128, opened 2022-01-04
Update footer copyright date dynamically

Kukurijek:
See discussion here -> #1126


rogerluan (approved):
LGTM! Great improvement 🌟 
Thanks for digging into this @Kukurijek !
Tested and approved 💪

PR #1149, opened 2022-04-12
docs(ci): update macos runner name

hpierre74:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

Seems not to be a generated doc from what I checked 🤞


rogerluan (approved):
Thank you! Nice catch! 🤗

PR #1162, opened 2022-07-02
Refer to built-in status bar modification method

ZevEisenberg:
`override_status_bar` is built in, and uses a first-party supported method for overriding the status bar, so the docs should probably mention that. If we want to refer to SimulatorStatusMagic for more advanced overrides, I suppose we could, although I'm not sure if it's supported any more.


rogerluan (approved):
I like this! Thanks for documenting it 🤗

PR #1164, opened 2022-07-04
Fix failing Netlify preview builds

janbrasna:
Trying to find a way for Netlify to acknowledge any PYTHON_VERSION attempts…

- [x] Pipfile — NO, no change:/
- [x] netlify.toml — NO, no context would apply the env even when correctly picking up the definitions…
- [ ] (don't have access to netlify UI if any change set there would make a difference… that's for @joshdholtz to try)
- [x] runtime.txt — NO, doesn't read the correct `python-3.7.y` format
- [x] runtime.txt — YEA, only when used in a weird `3.7` format…


janbrasna:
Nope:/

```
4:06:51 PM: Installing dependencies from Pipfile
4:06:54 PM: Warning: Python 3.7 was not found on your system...
4:06:54 PM: Neither 'pyenv' nor 'asdf' could be found to install Python.
4:06:54 PM: You can specify specific versions of Python with:
4:06:54 PM: $ pipenv --python path/to/python
4:06:54 PM: Error installing Pipenv dependencies
4:06:54 PM: Please see https://github.com/netlify/build-image/blob/xenial/included_software.md for current versions
4:06:54 PM: Build was terminated: Build script returned non-zero exit code: 1
```

janbrasna:
Nope, the env from netlify.toml doesn't get reflected either:/ even thou it should override any project-level settings in netlify admin — maybe setting the PYTHON_VERSION directly there in the UI would work.

The only working way was a nonstandard runtime.txt (which in turn may break others toys like docker that rely on the correct format) that I don't really fancy:/

janbrasna:
`deploy/netlify` — Deploy Preview ready! https://deploy-preview-1164--fastlane-docs-preview.netlify.app

> Screenshot 2022-07-04 at 19 49 18

So… at the moment the only working way is to add back the runtime.txt that was here before, like in #603 & then got removed in #1051 _(not sure why, OTOH to get it working in xenial you'd have to stick with versions they have provided, where 3.7 is the only viable TBH)…_ while an intent to make it work via env settings #1121 failed like my attempts.

Don't know if this "malformed" runtime.txt is the right way, but I wasn't able to find any workaround.

janbrasna:
Aight so if anyone's OK with the `runtime.txt` stating `3.7` I'll either rebase/squash all the WIP that I've left behind intact for now as an evidence for inspection what has been tried… or the PR can be itself squash-merged to get rid of the WIP… or if you feel like closing this PR for posterity & I'll open a new one, with just the final commit?

revolter:
> maybe setting the PYTHON_VERSION directly there in the UI would work.

From what I remember, I did try this too, and it didn't work.

revolter (approved):
Thanks ❤️

rogerluan:
Thank you so much @janbrasna!! 🤩

PR #1165, opened 2022-07-06
Fix «Edit in GitHub» whitespace

janbrasna:
1. Fixes whitespace generated inside the `a href` value to make the URL 100% precise (as the extra space simply doesn't belong there) that appeared after the actions conditional edit link.
_(works better for both fastlane/docs and fastlane/fastlane links:)_
>  Screen Shot 2024-03-17 at 21 19 05

_(I might have gotten the whitespace controls be a bit too hungry, but better be safe than sorry…)_

2. Adds the proper space between the octoicon and the link text. (The default theme has it spaced correctly, but some of the custom styling made it impossible for me to correct this, so just added a plain old space there…)

> Screen Shot 2024-03-17 at 21 31 31


KrauseFx (approved):

rogerluan (commented):
Great change with attention to detail, thank you ❤️

PR #1169, opened 2022-07-14
Add a "Fully automatic" workflow

sowenjub:
This follows a discussion with @joshdholtz on Twitter.

I wanted to make sure I would still be able to upload builds from Xcode if fastlane stopped working for some reason.
But after reading the docs, my setup asked me for the git repo with the certificates, which startled me.
Also, I hit another roadblock because my release scheme wasn't shared.

### DIFF
--- a/docs/codesigning/getting-started.md
+++ b/docs/codesigning/getting-started.md
@@ -53,6 +53,19 @@ That's why it is recommended to specify a specific provisioning profile somehow:
 
 The instructions below are more advanced, and take more time to setup than other approaches.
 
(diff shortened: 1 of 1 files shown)


iBotPeaches:
Sorry for mega delay. Looking at this its just a build action with a specified scheme and an upload - which I don't think belongs in the code signing area. Since its less related to Xcode versions and more about connecting dots.

I'd agree with you that we probably need a generic page with a full run through of both Android or iOS. The problem with the full run is it depends highly on the CI system used.

PR #1179, opened 2022-09-05
AWS Secrets manager storage

stanogn:
Could not find the generated docs source in the Fastlane repo so I'm just putting it here for reference.


iBotPeaches:
Unfortunately this will never work as it will get blown out on next release. A few of these docs are hard to find the original area. This one is at: https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/docs/sync_code_signing.md

Or `lib/fastlane/actions/docs/sync_code_signing.md`. Once changed in fastlane/fastlane - the next sync to here will fix it. If i merge this, it'll just be removed on next merge.

Sorry for delay.

PR #1182, opened 2022-09-26
fix dependency issues by updating Ruby & Python env

rogerluan:
## Description

fastlane main repo is having dependency issues to generate docs:

image

image

I think there's no need for us to stick with an older version of Ruby (a unspecified one) so maybe we can upgrade it already.

If this PR passes CI, it's good to go. However, I have a strong feeling that we'll also need to update the Ruby & Python versions installed in the fastlane docker image 😥 

----

I confirmed that this integration works with fastlane by testing it locally (at least locally, it works):

image

I modified this line locally:

https://github.com/fastlane/fastlane/blob/272063487b02b52510bb83172827fcbc336f4a55/fastlane/Fastfile#L518

To:

```ruby
sh("git clone #{git_url} --depth=1 --branch rogerluan-fix-dependency-issues-by-updating-everything")
```


rogerluan:
Oooh, it's so clever! It installs everything needed 😇 

image

rogerluan:
I need a review to move this forward 🙇 it's blocking PRs' GitHub checks in _fastlane_ main repo

.ruby-version
```diff
@@ -0,0 +1 @@
+3.1.1
```
ainame, line 1: Will 3.1.2 work? That's the latest version🙂

ainame (approved):

.ruby-version
rogerluan, line 1:
```suggestion
3.1.2
```

PR #1183, opened 2022-09-28
downgrade danger to keep supporting Ruby 2.6

rogerluan:
Turns out that https://github.com/fastlane/docs/pull/1182 didn't solve the problem when running `fastlane verify_docs` from Ruby 2.6 in CI, because the Ruby version used in `bundle install` within a lane is the Ruby version used to initiate the first process, and not the Ruby version specified in the current directory 😞 

This should revert the major version bump introduced here: https://github.com/fastlane/docs/pull/1181

This PR also downgrades the Ruby version from 3.1.2 to 3.1.1 because specifically on 3.1.2 there's an intermittent caching issue that causes the initial build on Netlify to work, but subsequent ones don't, until the cache is busted. Read more at: https://answers.netlify.com/t/cant-deploy-with-ruby-3-1-2/55728/9 @ainame 🙇


ainame (approved):
Oh no, thanks🙇

PR #1184, opened 2022-10-24
fix: cleanup description for changing project on google-credentials.md

bpeltonc:
This seeks to clarify some confusing parts of the setup docs for Android. Users may have setup the id for the linked Google Cloud project to be something that does not include the **Developer ID** from the Developer Console. This leaves the user guessing about which Google Cloud project should be selected when setting up a new Service Account, since they may not be able to find anything on the Google Cloud platform that references the Developer Id from the Developer console. The name displayed in the linked project on the API access page will always be one of the options in the Google Cloud project dropdown on that platform and changing the language in the docs this way eliminates any possible confusion there.


iBotPeaches (approved):

iBotPeaches:
thanks! sorry for delay. I agree that looking at URL, etc was a convoluted way to confirm when you can look at picker.

PR #1188, opened 2022-11-27
Add screencaps for permissions step

sitting-duck:
- Added screencaps for adding permissions for the Service account. 
- Removed the step for clicking the "Invite User" button as this button does not seem to exist anymore.


iBotPeaches:
thanks!

The problem I think we will have with screenshots is the UI changes near every 6 months and the job to obtain shots, redact image and re-publish is too much. We barely keep the text up to date.

On top of it - these images are just hotlinks to something. We'd need them in the `docs/img` folder. Sorry for mega delay but I'll close this out.

PR #1193, opened 2023-03-19
Add example of using a hash in App Store Connect API key docs

bartekpacia:
I think that section was lacking an example. It took me some time to Google the right syntax.

### DIFF
--- a/docs/app-store-connect-api.md
+++ b/docs/app-store-connect-api.md
@@ -92,6 +92,19 @@ end
 Keys and values that can be used in hash in `api_key` parameter in actions (e.g. `upload_to_testflight`) and tools (e.g. `cert`) are described in _fastlane_ API Key JSON file format below.
 The only difference when using hash is that you could use `filepath` instead of `key`.
 
+Below is an example of API key being used with CLI:
+
+```no-highlight
+$ bundle exec fastlane match appstore \
+    --api_key "
+        {
+            \"filepath\": \"/Users/bartek/Downloads/AuthKey_TJP2GMDZCI.p8\",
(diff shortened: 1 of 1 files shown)


docs/app-store-connect-api.md
```diff
@@ -92,6 +92,19 @@ end
 Keys and values that can be used in hash in `api_key` parameter in actions (e.g. `upload_to_testflight`) and tools (e.g. `cert`) are described in _fastlane_ API Key JSON file format below.
 The only difference when using hash is that you could use `filepath` instead of `key`.
 
+Below is an example of API key being used with CLI:
+
+```no-highlight
+$ bundle exec fastlane match appstore \
```
ainame, line 98: You don't have to break a line with `\` here but 👍

docs/app-store-connect-api.md
```diff
@@ -92,6 +92,19 @@ end
...
+    --api_key "
+        {
+            \"filepath\": \"/Users/bartek/Downloads/AuthKey_TJP2GMDZCI.p8\",
+            \"key_id\": \"TJP2GMDZCI\",
+            \"issuer_id\": \"57246542-96fe-1a63-e053-0824d011072a\"
```
ainame, line 103: To double-check, are they valid values? 😄 We can't access the API without the AuthKey file but make sure nothing bad will happen by using dummy values or invalidating the above key_id/issuer_id.

ainame (approved):

docs/app-store-connect-api.md
bartekpacia: These values are not real :)

PR #1198, opened 2023-06-13
Adding documentation for pure swift package unit testing

tmaly1980:
The docs for unit testing pure swift packages was not clear, this adds some use cases to make it more obvious.

NOTE: I just noticed the comment about saying this is not the correct location to submit a PR since it's autogenerated, but I cannot find the proper location. I cannot find run_tests anywhere in the docs repo and I don't see the markdown file anywhere in the primary fastlane repo. Please advise.

### DIFF
--- a/docs/generated/actions/run_tests.md
+++ b/docs/generated/actions/run_tests.md
@@ -237,7 +237,7 @@ Key | Description | Default
 ----|-------------|--------
   `workspace` | Path to the workspace file | 
   `project` | Path to the project file | 
-  `package_path` | Path to the Swift Package | 
+  `package_path` | Path to the Swift Package (ie "."), required if it's a pure Swift Package with no project or workspace file | 
   `scheme` | The project's scheme. Make sure it's marked as `Shared` | 
   `device` | The name of the simulator type you want to run tests on (e.g. 'iPhone 6' or 'iPhone SE (2nd generation) (14.5)') | 
   `devices` | Array of devices to run the tests on (e.g. ['iPhone 6', 'iPad Air', 'iPhone SE (2nd generation) (14.5)']) | 
--- a/docs/generated/actions/run_tests.md
+++ b/docs/generated/actions/run_tests.md
@@ -174,7 +174,7 @@ Returns | Outputs hash of results with the following keys: :number_of_tests, :nu
 
 
 
-## 6 Examples
+## 7 Examples
 
 ```ruby
 run_tests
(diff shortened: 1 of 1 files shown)


docs/generated/actions/run_tests.md
```diff
@@ -192,6 +192,18 @@ run_tests(
 )
 ```
 
+```ruby
+# run tests on a pure swift package (SPM), no xcodeproj or xcworkspace required
+run_tests(
+  package_path: ".",
```
rogerluan, line 198: Perhaps provide a more didactic example here? Like:

```suggestion
  package_path: "/path/to/package/directory" # The path to the directory where your Package.swift is in
```

docs/generated/actions/run_tests.md
```diff
@@ -192,6 +192,18 @@ run_tests(
...
+  scheme: "MySwiftCode-Package", # Must have -Package
```
rogerluan, line 199: Same here, e.g.:

```suggestion
  scheme: "MySwiftPackage-Package", # Append "-Package" to your Swift Package name
```

docs/generated/actions/run_tests.md
```diff
@@ -192,6 +192,18 @@ run_tests(
...
+  clean: true,
+  device: "iPhone 11 Pro", # Required for destination
+  result_bundle: true, # Last two lines may be required for proper xcresults output
+  output_directory: Dir.pwd + "/test_output"
```
rogerluan, line 203: Perhaps this would be better formatted like this?

```suggestion
  # These last two lines may be required for proper .xcresult output:
  result_bundle: true, 
  output_directory: Dir.pwd + "/test_output"
```

rogerluan (changes requested):
Hi @tmaly1980 👋  Thanks for improving fastlane docs!

This PR touches a file in the `/generated/` folder, which means it will be overwritten on the next deploy. You want to actually change code here: https://github.com/fastlane/fastlane/blob/0495da3504f8506587161c5ffb468f9694e65635/fastlane/lib/fastlane/actions/run_tests.rb#L93

Could you open a PR on fastlane main repo instead, updating the code above? When you do so, please close this one as it will be obsolete 🙏 

Thanks again!

iBotPeaches:
As Roger mentioned. This will not work as the next change of doc updates will blow out your changes. We need to adapt the file: `fastlane/lib/fastlane/actions/docs/run_tests.md` in fastlane/fastlane.

Once merged there - next release will have it here. So this change is meant for other repo. Sorry for confusion, I'll close this out.

PR #1211, opened 2023-09-18
Update runtime.txt for Netlify preview

joshdholtz:


janbrasna:
@rogerluan Yo, doesn't this by any chance align with moving the preview build distro from Xenial (legacy) to Focal (LTS) at Netlify? If so, the runtime.txt could be removed altogether (as 3.8 is /should be/ the default for Focal builds), see:
- old Xenial (PYTHON_VERSION, runtime.txt, Pipfile … 2.7 (default), 3.5, 3.7)
   - https://github.com/netlify/build-image/blob/xenial/included_software.md#languages
- vs.:
- new Focal (PYTHON_VERSION, runtime.txt, Pipfile … 3.8 (default), 2.7)
   - https://github.com/netlify/build-image/blob/focal/included_software.md#languages

The reason is the runtime.txt formatted this way for Netlify is not the correct syntax per se, see #1164 for elaboration.

Maybe now the LTS distro env would pick up all the other Python version definitions that were tried to no avail before on the old OS with no effect as demonstrated in #1121 etc.? Probably the current `Pipfile [requires] python_version = "3"` should alone be enough to make sure it builds on the new runner. (Why it was bugged on the old… who knows. But there were reports left to right running into the same bugs…)

rogerluan:
@janbrasna thanks for the suggestion, should this migration literally be just deleting the file, or does it require changing something in netlify?

Do you mind opening a PR for this so we can test? 😊

janbrasna:
@rogerluan My guess is yes, will dig around the logs to closely inspect what env is being picked up, TBD #1236 

(In the past, legacy image was used in Netlify admin UI and the hacks mentioned above were needed for the mkdocs to build. That distro has been deprecated and possibly completely removed and the project either manually or forcibly moved to the TLS Ubuntu images, where the hack is not needed anymore.)

PR #1216, opened 2023-10-08
add subresource integrity checks (SRI Hash) to some external resources

edent:
SRI is a new [W3C specification](https://www.w3.org/TR/SRI/) that allows web developers to ensure that resources hosted on third-party servers have not been tampered with. Use of SRI is recommended as a best-practice, whenever libraries are loaded from a third-party source.

Learn more about [how to use subresource integrity](https://developer.mozilla.org/docs/Web/Security/Subresource_Integrity) on MDN.

### DIFF
--- a/theme/base.html
+++ b/theme/base.html
@@ -23,7 +23,7 @@
   <link rel="stylesheet" href="{{ 'css/theme.css'|url }}" />
   <link rel="stylesheet" href="{{ 'css/theme_extra.css'|url }}" />
   {%- if config.theme.highlightjs %}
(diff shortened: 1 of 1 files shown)


rogerluan (commented):
This is great!

Could you share why the chosen algorithm was sha384 instead of sha512? And maybe document (in this PR's comments or description) where you got the sha from? I'm expecting you either generated it on your own, or copied it from an official source

Thanks in advance!

rogerluan (approved):
Thanks for your contribution @edent ! 🙌

PR #1225, opened 2023-11-09
fix `Match::Storage` docs to reflect latest API changes

vivek-dtdl:
Update instructions as Match::Storage changed in
https://github.com/fastlane/fastlane/pull/21018/


rogerluan (approved):
Thanks for bringing this to our attention @vivek-dtdl 🙏 and thanks for fixing it right away! 

LGTM 💪

PR #1236, opened 2024-02-24
Remove runtime.txt in favor of CI defaults

janbrasna:
Python 3.8 is the Netlify Focal build default now anyways, no need to force it.

> _— Originally posted by @janbrasna in https://github.com/fastlane/docs/issues/1211#issuecomment-1954843755:_
> 
> […] align with moving the preview build distro from Xenial (legacy) to Focal (LTS) at Netlify? If so, the runtime.txt could be removed altogether (as 3.8 is /should be/ the default for Focal builds), see:
> - old Xenial (PYTHON_VERSION, runtime.txt, Pipfile … 2.7 (default), 3.5, 3.7)
>    - https://github.com/netlify/build-image/blob/xenial/included_software.md#languages
> - vs.:
> - new Focal (PYTHON_VERSION, runtime.txt, Pipfile … 3.8 (default), 2.7)
>    - https://github.com/netlify/build-image/blob/focal/included_software.md#languages
> 
> The reason is the runtime.txt formatted this way for Netlify is not the correct syntax per se, see #1164 for elaboration.
> 
> Maybe now the LTS distro env would pick up all the other Python version definitions that were tried to no avail before on the old OS with no effect as demonstrated in #1121 etc.? Probably the current `Pipfile [requires] python_version = "3"` should alone be enough to make sure it builds on the new runner. (Why it was bugged on the old… who knows. But there were reports left to right running into the same bugs…)

Disclaimer: Can't predict it won't be needed at some point in future, but right now is not needed. Let's hope in the future other ways for the image to use correct env would suffice, it was the old image that was quite buggy. Only time will show how's the new image happy in applying the configs.

As I was the one adding back the runtime for python3 env under xenial to fix previews I didn't feel too comfortable leaving it long-term in this non-standard format anyways, so I'm more than happy to finally be able to remove it now.
(This was known back in #1164 that once moved to LTS preview image runner things would start to work without forcing the runtime… so it's time to un-shim it;)


janbrasna:
@rogerluan Can't seem to get to Netlify logs on passing builds (can access them only on failed, maybe you'd see more), but generally… it passes;)

Screen Shot 2024-02-24 at 19 36 43

CircleCI is happy too, picking up 3.8.2:

Screen Shot 2024-02-24 at 19 30 04

KrauseFx (approved):

PR #1242, opened 2024-03-17
Add build date to footer

janbrasna:
Adds `Y-m-d` last build date to footer (while removing some default bits that are more confusing than not /e.g. the sphinx template reference/…) to give an idea when master was last deployed to production.

Screen Shot 2024-03-17 at 17 34 05

(proposed compared to current)

### DIFF
--- a/theme/footer.html
+++ b/theme/footer.html
@@ -22,7 +22,7 @@
     </p>
   </div>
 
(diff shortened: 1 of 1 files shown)


KrauseFx (approved):
Excellent idea

rogerluan (commented):
💯 💯 💯

PR #1244, opened 2024-04-05
[WIP] Try CI without .ruby-version

janbrasna:
TL;DR it basically only confuses Netlify, so if we can do without it …


janbrasna:
This is just sad, I have the same PR in my fork running the exact preview build image (check the sha for laughs), and the result is different Ruby version ;DD and a failure based on that. The only difference really is: caches.

Screenshot 2024-04-05 at 19 29 46

PR #1251, opened 2024-06-29
Fix broken link and add more detailed instructions in the App Store Connect API page

Raja0sama:
They renamed Keys to Integration, which miss lead me for a while, until I realize that the documentation is pointing me in the wrong direction.

### DIFF
--- a/docs/app-store-connect-api.md
+++ b/docs/app-store-connect-api.md
@@ -36,7 +36,7 @@ Below are the statuses of each tool:
 
 1. Create a new App Store Connect API Key in the [Users page](https://appstoreconnect.apple.com/access/api)
   - For more info, go to the [App Store Connect API Docs](https://developer.apple.com/documentation/appstoreconnectapi/creating_api_keys_for_app_store_connect_api)
-  - Select the "Keys" tab
+  - Open the ‘Integrations’ tab, select ‘App Store Connect API,’ and create a new TEAM API using the ‘+’ button. 
   - Give your API Key an appropriate role for the task at hand. You can read more about roles in [Permissions in App Store Connect](https://developer.apple.com/support/roles/)
   - Note the Issuer ID as you will need it for the configuration steps below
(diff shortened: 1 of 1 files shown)


docs/app-store-connect-api.md
```diff
@@ -36,7 +36,7 @@ Below are the statuses of each tool:
 
 1. Create a new App Store Connect API Key in the [Users page](https://appstoreconnect.apple.com/access/api)
```
rogerluan, line 37:
```suggestion
1. Create a new App Store Connect API Key in the [Users and Access](https://appstoreconnect.apple.com/access/users) page
```

docs/app-store-connect-api.md
```diff
@@ -36,7 +36,7 @@ Below are the statuses of each tool:
...
   - For more info, go to the [App Store Connect API Docs](https://developer.apple.com/documentation/appstoreconnectapi/creating_api_keys_for_app_store_connect_api)
-  - Select the "Keys" tab
+  - Open the ‘Integrations’ tab, select ‘App Store Connect API,’ and create a new TEAM API using the ‘+’ button.
```
rogerluan, line 39:
```suggestion
  - Open the `Integrations` tab, select `App Store Connect API`, and create a new "Team Key" using the `+` button
    - Note that if you can't see the `App Store Connect API` tab, this means you don't have permission yet. Please refer to the docs above to know how to get this permission
```

rogerluan: Normally I prefer ending every sentence with a period, but the pattern here is to not have an ending period so I opted for consistency 🤕

rogerluan (changes requested):
Nice spotting! Just some minor formatting improvements and fixing a broken link as well

rogerluan (approved):
💪 Thanks for bringing this to our attention, @Raja0sama ! 🚀

PR #1254, opened 2024-07-13
Restore GA tracking

### LINKED ISSUE #1253: GA4 may continue working using legacy UA setup
In #1213 updating mkdocs & dep versions a new GA/GTM setup needed to be used. Without a proper GA4 migration all the previous GA3 config was omitted under the impression it can't be gathering data anymore.

Fortunately Alphabet migrated GA3 UA properties to GA4 for those who never migrated it themselves. https://support.google.com/analytics/answer/11583528#zippy=%2Cwhat-happens-to-my-ua-connected-site-tag :
> "For Google Analytics 4 properties that rely on Universal Analytics tagging for data collection, there is no additional risk when those Universal Analytics properties are shut down and deleted by Google. Once Google deletes your Universal Analytics property we’ll automatically continue to use your Universal Analytics tagging for your Google Analytics 4 property, however only if your UA tags are still on your website. If you decide to remove your UA tags, you can replace them with a Google tag."

There's even continued (limited) support for legacy analytics.js:

> GA3-UA-GA4

The template here overrides the defaults providing its own GA snippet based on the old analytics.js anyways, so this legacy path is still valid, just needs reconfiguring. (Expecting the GA3 being auto-migrated to GA4 and the analytics.js ingestion enabled by default as a result.)

janbrasna:
GA4 integration should continue gathering data for (auto)migrated GA3/UA properties.

Resolves #1253

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -7,6 +7,8 @@ repo_url: https://github.com/fastlane/docs
 theme:
   name: readthedocs
   custom_dir: theme
+  analytics:
+    gtag: UA-18658848-13
 strict: false
 markdown_extensions:
 - markdown.extensions.attr_list
--- a/theme/base.html
+++ b/theme/base.html
@@ -56,14 +56,14 @@
   {%- block extrahead %} {% endblock %}
 
   {%- block analytics %}
-  {% if config.google_analytics %}
+  {% if config.theme.analytics %}
   <script>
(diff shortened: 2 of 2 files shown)


janbrasna:
@rogerluan This is sorta emergency fix, could you please look into publishing this as you get to it? Thanks! — Sorry for misleading you in the first place!

PR #1256, opened 2024-07-31
Fix xcode version for circleci

cprodhomme:


docs/best-practices/continuous-integration/circle-ci.md
```diff
@@ -55,7 +55,7 @@ orbs:
 jobs:
   build:
     macos:
-      xcode: "13.1.0"
+      xcode: "15.4.0"
```
janbrasna, line 58: Currently [v2.222 tests](https://github.com/fastlane/fastlane/blob/e1ab951b63aeb7a669b415708ffbe0e1346c0f59/.circleci/config.yml#L282-L285) agiainst 15.3.0 so these should probably be kept in sync.

janbrasna: (i.e. once 15.4 is also added there I'm all for having the most uptodate here, given the examples have no ruby version constraints too that would make the CircleCI env angry…)

iBotPeaches:
Sorry for mega delay on this. With a few gone by - Xcode 26 is the way, but this script is kinda old now - not using orbs, etc. Closing this out for a new PR I did to correct both - https://github.com/fastlane/docs/pull/1321

PR #1276, opened 2025-03-06
Enhance example of app store connect api usage with actions

giuliobracci:
Hello,

First of all, thank you for the great project.

I encountered some difficulties following the provided examples for setting up App Store Connect API authentication. This was mainly because I was unfamiliar with the `pilot` action/tool and mistakenly believed that I needed to invoke it when, in fact, it wasn’t required. Given Fastlane’s extensive options and features, I believe adding a small comment for clarification could improve usability, especially for users less familiar with the Fastlane ecosystem.

### DIFF
--- a/docs/app-store-connect-api.md
+++ b/docs/app-store-connect-api.md
@@ -65,7 +65,7 @@ lane :release do
     in_house: false # optional but may be required if using match/sigh
   )
 
-  pilot(api_key: api_key)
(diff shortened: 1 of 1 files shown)


iBotPeaches:
thanks! I don't see much benefit of a comment after the action. I acknowledge the difficulties, but I don't think this is the way.

Issue #1278, opened 2025-03-29
`Available Plugins` content is outdated


ntsk:
https://docs.fastlane.tools/plugins/available-plugins/

![Image](https://github.com/user-attachments/assets/f991ec75-4d22-45b0-aa89-df4087958967)

It seems that this page is auto-generated, but its content has not been updated.

I believe the HTML is generated from the following Markdown file, but this file hasn't been updated for about four years:
https://github.com/fastlane/docs/blob/master/docs/generated/available-plugins.md

On the other hand, this Markdown file appears to be relatively new:
https://github.com/fastlane/docs/blob/master/docs/plugins/available-plugins.md

iBotPeaches:
That took a bit to figure out - fastlane/fastlane caused it, but I fixed up docs one-time and changed some files around to prevent it from happening again.

 - https://github.com/fastlane/docs/pull/1287
 - https://github.com/fastlane/fastlane/pull/29729

PR #1287, opened 2025-10-19
fix: correctly update the generated available-plugins during deploys.

iBotPeaches:
## The Problem

:key: We see on the public docs that available plugins are outdated severely. Roughly 4 years.

This is because files that are generated are supposed to be stored in `docs/generated/` and then a script during mkdocs build copies those files out of generated into the final resting place.

```yaml
- mkdocs-simple-hooks:
    hooks:
      on_pre_build: hooks:copy_generated_files
```

That script copies 2 files and a directory prior to doc build.

```python
def copy_generated_files(*args, **kwargs):
    shutil.copy("docs/generated/available-plugins.md", "docs/plugins/available-plugins.md")
    shutil.copy("docs/generated/actions.md", "docs/actions.md")
    
    actions_dir="docs/actions"
    if os.path.isdir(actions_dir):
        shutil.rmtree(actions_dir)
    shutil.copytree("docs/generated/actions", actions_dir)
```

So every release we copy a file that has not been updated in years overwriting the one that was freshly updated via the release. So the "Available Plugins" has been frozen in time for 4 years.

This requires a fix in both repositories since the bug occurred when fastlane/fastlane started writing files not in the generated folder.

## The Fix

We git-ignore files in the main tree that are built via automation. This will prevent humans from changing files that should not be changed and push us to the automation/generated path.

This PR on fastlane/fastlane side will fix it going forward - https://github.com/fastlane/fastlane/pull/29729

fixes: #1278


iBotPeaches:
My preview here shows the up to date available plugins for context

new - https://deploy-preview-1287--fastlane-docs-preview.netlify.app/plugins/available-plugins/
old  - https://docs.fastlane.tools/plugins/available-plugins/

PR #1290, opened 2025-11-06
[ci] Migrate CircleCI to Ruby 3.3, fix CI

iBotPeaches:
## Problem
The CI platform on docs is built from an outdated [DockerHub account](https://hub.docker.com/r/fastlanetools/fastlane/) only in the hands of Josh. I tried for about a year to get keys/access to that account, but failed. This meant as we moved fastlane/fastlane to new minimum Ruby versions, docs fell behind unable to be built/deployed.

## Solution
I've created a new DockerHub account and published images with full automation [here](https://github.com/fastlane/docker/releases/tag/v1.0.1). I've ported the changes from fastlane/fastlane in terms of CI (https://github.com/fastlane/fastlane/pull/29753) to bring this CI pipeline to modernization.


iBotPeaches:
Hmmm. So the challenge here is you need Python (mkdocs) and Ruby (fastlane) and doing that with native Docker images is a challenge.

I will return to the fastlane/docker repo to bring an updated Ruby 3.x image that includes Python 3.8 and then setup some CI/CD for automated deploys to DockerHub. Since it seems that image is really dependent on this setup.

Alternatively with GitHub Actions, we could just install Ruby/Python with setup-python/setup-ruby, but extracting all the secrets and flows from this working CircleCI would present a challenge.

iBotPeaches:
PR opened on docker - I may test with my forked image here in meantime - https://github.com/fastlane/docker/pull/11

.circleci/config.yml
```diff
@@ -1,29 +1,30 @@
...
+      - python/install-packages:
+          pkg-manager: pipenv
+      - ruby/install-deps
       - run: pipenv run bundle exec fastlane test
       - run: bundle exec rubocop
       - run: bundle exec danger || echo "danger failed, moving on"
   "Deploy":
     docker:
-      - image: fastlanetools/ci:0.4.0
+      - image: ibotpeaches/fastlaneci-test:latest
```
rogerluan: Should this point to our image?

iBotPeaches: Good catch. Aligned it in latest commit

taquitos (approved):
LGTM!

PR #1302, opened 2026-01-08
chore: slack_train is a plugin with actions, not official actions

iBotPeaches:
The removal logic for auto-generated actions is not the best. It will keep them up to date, but in the sense of removal - the file will never be updated on fastlane/docs side.

So when the PR comes across during doc updates (https://github.com/fastlane/docs/pull/1299/changes) it builds on top of the previous docs but not with removals.

So I'm under impression that I can remove the outdated slack_train action stuff here, because we already have the slack_train plugin being parsed.


lacostej (approved):
Is there an issue somewhere about the fact that the removal isn't working automatically?

iBotPeaches:
> Is there an issue somewhere about the fact that the removal isn't working automatically?

made one. I did not see one - https://github.com/fastlane/docs/issues/1317

PR #1309, opened 2026-02-27
docs: add note about 2.7 min and 3.3+ preferred

iBotPeaches:
fixes: #1305

PR #1310, opened 2026-03-03
Bump required Ruby version up to 2.7

kaibolay:
also mention that Ruby >= 3.2 is recommended


kaibolay:
d'oh! #1309 already exists

PR #1319, opened 2026-04-19
Update docs for latest fastlane release 2.233.0 (actions.md, available-plugins.md) 🚀

iBotPeaches:
Release v2.233.0


iBotPeaches:
Too greedy of an upgrade. Took bundle with it. I'll patch fastlane/fastlane and redo

PR #1324, opened 2026-04-25
build: migrate to uv, drop pip

iBotPeaches:
## Problem

Its hard to work on these docs. The decaying nature of the setup means it takes some work to get all the pieces in order just to spawn local docs.

## Solution
Drop pipfile, move to uv and make this simple. Now you just `uv sync` and `uv run mkdocs serve` - done.


netlify.toml
```diff
@@ -1,4 +1,4 @@
 [build]
-command = "mkdocs build"
+command = "curl -LsSf https://astral.sh/uv/install.sh | sh; uv sync; uv run mkdocs build"
```
lacostej, line 2: I never like using those curl | sh commands in automated setups, but is there anything better for `uv`?

lacostej (approved):

netlify.toml
iBotPeaches: In about 6 months it's rumored Netlify will native UV support. I'll find the link

iBotPeaches: Guess it wasn't a planned timeline - https://answers.netlify.com/t/feature-requests-what-do-you-already-love-what-could-be-better/256/236

janbrasna: Seems they have neither pipx preinstalled, nor brew available… given it's an ephemeral runner so the package mess is not an concern, could just 
```
pip install uv; uv sync; uv run mkdocs build
```
handle the same? (not sure if exposes things as expected that way though)

iBotPeaches: Maybe thats a bit safer using pip to bootstrap uv vs blindly shelling a URL. Maybe Netlify supports UV natively at this point

iBotPeaches: Netlify still doesn't natively support uv so did this - https://github.com/fastlane/docs/pull/1341

PR #1339, opened 2026-06-28
docs: Ruby 3 is min now

iBotPeaches:
As of fastlane 2.235.0 - Ruby 3 is our min - https://github.com/fastlane/fastlane/releases/tag/2.235.0


rogerluan (approved):
🚀

PR #1341, opened 2026-07-01
chore: install uv via pip

iBotPeaches:
## Problem

It's a bit insecure to CURL and then just shell a random online file. 

## Solution

I believe we can trust Astral, but more secure until Netlify natively supports "uv" we can use pip to install uv like recommended here - https://docs.astral.sh/uv/getting-started/installation/#pypi

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,4 +1,4 @@
 [build]
-command = "curl -LsSf https://astral.sh/uv/install.sh | sh; uv sync; uv run mkdocs build"
+command = "pip install uv; uv sync; uv run mkdocs build"
 publish = "site/"
 environment = { PYTHON_VERSION = "3.8.13" }


janbrasna:
LGTM: https://app.netlify.com/projects/fastlane-docs-preview/deploys/6a4a86ac2236420008e6ffdf#L42

```sh
$ pip install uv; uv sync; uv run mkdocs build
Requirement already satisfied: uv in /opt/buildhome/.local/share/mise/installs/python/3.8.13/lib/python3.8/site-packages (0.11.26)
Resolved 27 packages in 11ms
Checked 25 packages in 8ms
```

(while it appears as `uv` got included during some bootstrapping already, this is definitely a reasonable way to either make sure it's available or get it installed — logs look happy.)

rogerluan (approved):

Issue #974, opened 2020-09-13
Dark mode


bartekpacia:
Did you consider creating the dark version of the website?

janbrasna:
Not sure how much of the underlying RTD template is still actively used here, but there's a long running upstream issue: https://github.com/readthedocs/sphinx_rtd_theme/issues/224 — with a couple smart hacks as https://github.com/readthedocs/sphinx_rtd_theme/issues/224#issuecomment-722655021 and/or reference to a separate implementation as https://github.com/MrDogeBro/sphinx_rtd_dark_mode or alternative https://github.com/jucacrispim/sphinx_pdj_theme that might serve as inspiration how involved this might be.

janbrasna:
(There used to be an attempt to move from RTD template to Material template, and that one indeed does have a dark theme: https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/ … however that move was cut short at some point as the template work for mkdocs upgrade was based on the current original that was kept.)

janbrasna:
The actual design is quite similar to what Furo theme https://github.com/pradyunsg/furo has out-of-the-box, and that one DOES HAVE the dark theme functionality. Maybe that could be the upgrade path?

Brett-Best:
I've opened draft PR [#1343](https://github.com/fastlane/docs/pull/1343) as an explicitly AI-driven experiment to explore dark mode as an accessibility improvement.

It keeps the customized Read the Docs theme rather than reviving the Material migration from #696. MkDocs' native color-mode controls only support its built-in `mkdocs` theme, while this site relies heavily on RTD-specific templates, navigation, Algolia search, and several thousand lines of custom CSS. The draft therefore needs a small persisted `Auto` / `Light` / `Dark` controller plus targeted styling for code, tables, admonitions, search, plugins, mobile navigation, branding, and print.

The experiment includes desktop/mobile/search/print browser checks and an APCA comparison against the current light theme. The comparison found contrast parity and no identified dark-mode regression, but this is not a claim of WCAG or APCA compliance. The PR documents the remaining contrast clusters and implementation tradeoffs, and I would especially welcome human accessibility and design review before it is considered ready.

PR #1343, opened 2026-07-12
feat: add dark mode

Brett-Best:
> [!IMPORTANT]
> This draft is explicitly an AI-driven experiment exploring dark mode as an accessibility improvement. The implementation, tooling research, browser checks, APCA comparison, and initial code review were AI-assisted. It is not a claim of WCAG or APCA compliance, and it should receive human accessibility and design review before being considered ready.

## Problem

Addresses #974.

The docs currently provide only a light appearance. Dark mode can reduce discomfort for people who are sensitive to bright interfaces or who browse in low-light environments, but adding it here is more involved than enabling a MkDocs setting.

## Why this needs a custom approach

[MkDocs' native `color_mode` and `user_color_mode_toggle` options](https://www.mkdocs.org/user-guide/choosing-your-theme/#mkdocs) apply only to its built-in `mkdocs` theme. This site uses a substantially customized `readthedocs` theme instead, including a forked base template, RTD-specific `.wy-*` and `.rst-content` styling, custom navigation, Algolia DocSearch, plugin-quality pages, and project-specific components.

The earlier discussion in #974 and the unfinished Material-theme migration in #696 illustrate the tradeoff: switching themes to gain a built-in toggle would mean reworking the site's existing templates and thousands of lines of theme-specific CSS. It would be a broad redesign rather than a contained accessibility experiment.

This draft therefore keeps the current theme and adds color-mode support around its existing structure. That requires handling several pieces explicitly:

- apply the saved or system preference before styles paint, avoiding a light-mode flash
- persist an `Auto` / `Light` / `Dark` choice and react to live system changes
- cover RTD navigation, content, forms, tables, admonitions, code and syntax highlighting
- load DocSearch styles early enough for local dark-mode overrides, and style its legacy v2 dropdown
- account for custom plugin scores, setup panels, responsive navigation, and the fastlane wordmark
- keep printed documentation light and hide the interactive selector

## What changed

- adds an accessible color-mode selector to the sidebar
- resolves the initial preference in the document head, then manages persistence, system changes, and cross-tab updates with dependency-free JavaScript
- adds a semantic dark palette and targeted styles for the customized RTD theme
- preserves the colored fastlane mark while making the wordmark readable on dark surfaces
- keeps Algolia results on the current preview host so search can be exercised locally
- ignores generated Python cache files and updates the homepage's legacy Twitter badge and link to X

No documentation routes or navigation entries were changed in `mkdocs.yml`; it only registers the new CSS and JavaScript assets.

## Accessibility and browser validation

This was evaluated as an experiment rather than certified as compliant:

- captured five representative pages in both light and dark modes using Playwright at 1280px full-page resolution
- exercised desktop, mobile navigation, persisted selection, live `prefers-color-scheme` changes, Algolia search, code blocks, tables, admonitions, plugin-quality pages, and print rendering
- observed no browser console errors or warnings in the tested flows
- ran an exploratory APCA-W3 (4g) audit over roughly 250–470 visible text elements per page using bronze-level thresholds

The APCA comparison found contrast parity with the existing light theme: 17 unique failing dark-mode style clusters versus 14 pre-existing light-mode clusters, with a similar minimum contrast around Lc 51. The remaining dark clusters are primarily muted secondary text, GitHub's shipped dark syntax palette, and link colors; their light-theme counterparts also fall below the same bronze thresholds. No dark-mode regression was identified, but these results should not be interpreted as formal compliance.

## Repository checks

- `bundle exec fastlane test`
  - strict MkDocs build
  - filename, redirect, image, and asset-location checks
  - all code samples from 238 files
- `node --check docs/js/color-mode.js`
- `git diff --check upstream/master...HEAD`

## Maintenance notes

The small inline head script intentionally duplicates the storage key and valid preference values from `color-mode.js`; this prevents a flash of the wrong theme, but those values must remain synchronized.

The CSS is deliberately comprehensive because it sits on top of a heavily customized legacy theme. A possible follow-up would extract more of the existing light-theme colors into semantic tokens and use dedicated per-mode syntax-highlighting stylesheets, which would reduce the number of targeted overrides over time.

Human review is especially welcome on accessibility priorities, palette choices, and whether this incremental approach is preferable to a future full theme migration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ site/
 *.swp
 vendor/bundle
 .venv
+__pycache__/
 
 # these files are generated during documentation build
 docs/actions.md
--- a/docs/css/color-mode.css
+++ b/docs/css/color-mode.css
@@ -1,30 +1,48 @@
-/* Color mode control */
-.color-mode-control {
+/* Color mode toggle */
+.wy-side-nav-search {
+  position: relative;
+}
+
+/* Hidden until the pre-paint script tags <html>, so visitors without
+   JavaScript are not offered a control that cannot work for them. */
+.color-mode-toggle {
+  display: none;
+}
+
+html[data-color-mode-preference] .color-mode-toggle {
   align-items: center;
-  display: flex;
-  font-size: 13px;
-  gap: 8px;
-  justify-content: center;
-  margin: 0 auto 4px;
-}
-
-.color-mode-control label {
-  color: #666;
-  font-weight: 600;
-  margin: 0;
-}
-
-#color-mode-select {
-  background-color: #fff;
-  border: 1px solid #6d6d6d;
-  border-radius: 3px;
-  color: #333;
+  background: transparent;
+  border: 0;
+  border-radius: 50%;
+  color: #555;
   cursor: pointer;
-  font: inherit;
-  padding: 3px 24px 3px 7px;
+  display: flex;
+  height: 32px;
+  justify-content: center;
+  padding: 0;
+  position: absolute;
+  right: 8px;
+  top: 10px;
+  width: 32px;
 }
 
-#color-mode-select:focus-visible,
+.color-mode-toggle:hover {
+  background-color: rgba(0, 0, 0, .08);
+  color: #111;
+}
+
+.color-mode-icon {
+  display: none;
+  height: 18px;
+  width: 18px;
+}
+
+html[data-color-mode-preference="auto"] .color-mode-icon-auto,
+html[data-color-mode-preference="light"] .color-mode-icon-light,
+html[data-color-mode-preference="dark"] .color-mode-icon-dark {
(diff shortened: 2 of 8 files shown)


iBotPeaches:
Tested it out - works pretty well. I guess since mkdocs is on the way out and probably has no short-term future. Probably  no big problem to really build further onto our iteration of it.

Maybe the color switcher can be a bit more elegant is my only 2c. It sits up there with a dropdown vs like an emoji switcher like I commonly see these days.

Screenshot 2026-07-15 at 7 01 44 AM

Brett-Best:
Updated to improve that + add a crossfade, https://deploy-preview-1343--fastlane-docs-preview.netlify.app

Given it deploys out to a Netlify preview I'm not in a rush to have this merged given it is straight up AI slop

PR #1344, opened 2026-08-13
docs: fix indents on getting started for Android

iBotPeaches:
## Problem

Screenshot 2026-08-13 at 6 26 47 AM

## Solution

Screenshot 2026-08-30 at 11 18 30 AM

### DIFF
--- a/docs/includes/google-credentials.md
+++ b/docs/includes/google-credentials.md
@@ -12,26 +12,34 @@
    1. Provide a `Service account name` (e.g. fastlane-supply)
    1. Copy the generated email address that is noted below the `Service account-ID` field for later use
    1. Click **DONE** (don't click **CREATE AND CONTINUE** as the optional steps such as granting access are not needed): <img src="/img/getting-started/android/creating-service-account.png" width="700" />
-1. Setup [authentication](https://cloud.google.com/iam/docs/best-practices-service-accounts#choose-when-to-use) for the service account, fastlane supports Workload Identity Federation and service account keys
-  - When using Workload Identity Federation (such as for GitHub Actions)
-    1. Create a Workload Identity Pool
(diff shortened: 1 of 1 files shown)


rogerluan (changes requested):
Indented bullets are rendering as regular unordered bullets but the md file is using ordered bullets. Should fixing this also belong to this PR?

iBotPeaches:
> Indented bullets are rendering as regular unordered bullets but the md file is using ordered bullets. Should fixing this also belong to this PR?

hmm I see your point. Probably not. I think its way better than shoved into 1 line as it stands now and the ordered/unordered must be some old bug in this outdated mkdocs stack

iBotPeaches:
Screenshot 2026-08-15 at 2 18 30 PM

Yeah odd @rogerluan - unordered breaks the formatting entirely, no matter how many indents or not I use. Using an ordered list fixes indent, but then renders as unordered.

rogerluan:
@iBotPeaches do you wanna investigate and fix that in this same PR, or separate?

iBotPeaches:
Never personally lol. Mkdocs is dead. This fixes an obvious regression of formatting, personally unordered or ordered does anyone mind?

iBotPeaches:
I started a migration to ProperDocs (fork of mkdocs) so we can hopefully start fixing some of these bugs - https://github.com/fastlane/docs/pull/1345

Without ingesting a massive amount of work moving to like Docusaurus or Zensical. A drop-in replacement that isn't abandoned may be best path forward to fix some of these odd bugs.

iBotPeaches:
@rogerluan - I reworked it entirely into 2 different auth sections to workaround nested bug. Lemme know what you think now, just skips the problem entirely now.

